PR #1, opened 2019-10-15
Move

christopherhans:


vrops-exporter.py
```diff
@@ -0,0 +1,95 @@
...
+
+        print(response)
+        json_response = response.json()
+        for resource in json_response["resourceList"]:
+            print(resource["identifier"], resource["resourceKey"]["name"])
+        print("Completed")
+
+        entityname = "vmentityname"
+
+        metric = Metric("vrops_virtualmachine", "virtualmachine from vrops", "gauge")
```
viennaa: Please use the classic way of [GaugeMetricFamily](https://github.com/prometheus/client_python/tree/6214c3a5b4badeab3fb45baa17a6edcd208f57e7#custom-collectors), add_sample [should not really be used](https://github.com/prometheus/client_python/blob/3cb4c9247f3f08dfbe650b6bdf1f53aa5f6683c1/prometheus_client/metrics_core.py#L38-L42).
Please also add a try block around it.

vrops-exporter.py
```diff
@@ -0,0 +1,95 @@
...
+                          value=1, labels={"target": self._target,
+                                           "entityname": entityname})
+
+        yield metric
+
+
+requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
+
+# main
+if __name__ == '__main__':
```
viennaa: The main function is the content I would like to see in here. Please enhance it, so that params can either be read via config file or cli option. Add option for debug.

Move the whole Collector part into a dedicated file and name it XxxCollector

module/read_yaml.py
```diff
@@ -0,0 +1,15 @@
...
+        self._path = path
+
+    def run(self):
+        yml = dict()
+        with open(self._path, 'r') as stream:
+            try:
+                yml = yaml.safe_load(stream)
+            except yaml.YAMLError as exc:
+                print(exc)
+        return yml
```
viennaa: Filename should start with an uppercase letter and since the class is named YamlRead I'd name the file YamlRead.py too.

christopherhans:
@viennaa done

PR #3, opened 2019-10-22
implemented auto detection for modules

christopherhans:
**Opened this PR for discussion.**
@viennaa @richardtief
I implemented a method which automatically imports and runs all classes from the module/ directory.
If someone wants to contribute to the vrops-exporter, he/she only needs to add a new module (which requests the metric via REST and transform it  into a prometheus metric), w/o knowledge about the overall exporter. The newly added module needs to be mentioned nowhere.

The design of the modules can be discussed.

Open for code improvements.


christopherhans:
Of course, that needs some constraints. Like modules have to base on the ABC etc...
But since this is a PR only repo, we can ensure this.

VropsCollector.py
```diff
@@ -51,37 +51,16 @@ def __init__(self, target):
...
+            self._modules_dict[module] = importlib.import_module(module, modules[0])
 
     def collect(self):
-        if os.environ['DEBUG'] == '1':
-            print('started')
+        for target in self._modules_dict.keys():
+            mod = self._modules_dict[target]
+            func = getattr(mod, target)
+            res = func(self._target, self._user, self._password).collect()
+            yield res
```
viennaa: if you are directly yielding the result, you will only be able to have a single metric per module.
Iterate over the results and yield every item.

BaseCollector.py
```diff
@@ -0,0 +1,7 @@
+from abc import ABC, abstractmethod
+
+
+class BaseCollector(ABC):
+
+    def collect(self):
```
viennaa: I'd make it an `@abstractmethod`

module/SampleCollector.py
```diff
@@ -0,0 +1,44 @@
...
+
+        if os.environ['DEBUG'] == '1':
+            print("Completed")
+
+        entityname = "vmentityname"
+
+        g = GaugeMetricFamily('vrops_ressource_gauge', 'Gauge Collector for vRops',
+                              labels=['target', 'entityname'])
+        g.add_metric(labels=[self._target, entityname], value=1)
+        return g
```
viennaa: like written earlier, better collect them in a list an return a list of metrics.

BaseCollector.py
christopherhans: My mistake, imported abstractmethod, but forgot to use it.

module/SampleCollector.py
christopherhans: now added to a list. In this case its only one metric, so it might look a little bit strange.

viennaa: If you feel like it, you could add another sample metric, to make it clearer. No absolute need imho.

PR #6, opened 2019-11-14
ResourceCollector

christopherhans:
*Do not merge!*

Created a directory `resources` which contains the resource classes.
Created the resource classes.
Created `get_resource` function.

To do: Inital `vcenter-uuid` and `datacenter-uuid` needs to be fetched before creating a `vcenter-object`.


resources/VirtualMachine.py
```diff
@@ -0,0 +1,6 @@
+class VirtualMachine:
```
viennaa:
```suggestion
class VirtualMachine(Host):
```

Adding the parent relation like this will be helpful to go up the tree later. Please do it for the other classes to up to Vcenter

PR #19, opened 2019-12-03
Resource on testing 2

viennaa:


richardtief:
make some wrapper methods make some wrapped methods and from the `get_resources` tool a `Resources` class to make the code testable

viennaa:
Looks promising, now you should be able to mock your test structure

richardtief:
small changes in the way the collection of project ids works. It works so far. Unfortunately the vrops in QA has no project IDs to test. 
The code structure is ready to test. That will be implemented next.

viennaa:
Test is still failing:

```
Traceback (most recent call last):
  File "tests/TestCollectors.py", line 15, in <module>
    from tools.Resources import get_resources
ImportError: cannot import name 'get_resources' from 'tools.Resources' (./tools/Resources.py)
```

resources/Cluster.py
```diff
@@ -1,17 +1,17 @@
-from tools.get_resources import get_resources
+from tools.Resources import Resources
```
viennaa: just import tools.Resources

resources/Cluster.py
```diff
@@ -1,17 +1,17 @@
...
         self.hosts = list()
 
     def add_host(self):
-
-        for hosts in get_resources(self, resourcetype='resources', resourcekind='HostSystem'):
-            self.hosts.append(Host(hosts['name'], hosts['uuid']))
-
-
+        res = Resources()
+        for hosts in res.get_hosts(target=self.target, parentid=self.uuid):
```
viennaa: Call it directly with Resources.get...

resources/Datacenter.py
```diff
@@ -0,0 +1,16 @@
+from tools.Resources import Resources
```
viennaa: same as in Cluster.py

resources/Host.py
```diff
@@ -1,40 +1,23 @@
-from tools.get_resources import get_resources
+from tools.Resources import Resources
```
viennaa: same as in Cluster.py

resources/Host.py
```diff
@@ -1,40 +1,23 @@
...
-                                resourcekind="VirtualMachine"):
-            for vm_id in self.vms:
-                if vm['uuid'] != vm_id.uuid:
-                    self.vms.append(VirtualMachine(vm['uuid'], project_id='default internal'))
-
+        res = Resources()
+        project_ids = res.get_project_id(target=self.target)
+        for vm in res.get_virtualmachines(target=self.target, parentid=self.uuid):
+            if vm['uuid'] in project_ids:
+                print(vm['name'] + ' has project id: ' + project_ids['project_id'])
```
viennaa: make this print only happen if debug is set

tools/Resources.py
```diff
@@ -0,0 +1,60 @@
...
+from urllib3 import disable_warnings, exceptions
+from requests.auth import HTTPBasicAuth
+
+
+class Resources:
+
+    def get_resources(self, target, **kwargs):
+        url = "https://" + target + "/suite-api/api/resources"
+        querystring = {
+            'parentId': kwargs['parentid'],
```
viennaa: looks like you are relying on parentid and resourcekind here, provided as kwargs. You should rather explicitly expect them, otherwise this run will fail, won't it?

tools/Resources.py
```diff
@@ -0,0 +1,60 @@
...
+            'resourceKind': kwargs['resourcekind'],
+            'pageSize': '50000'
+        }
+        headers = {
+            'Content-Type': "application/json",
+            'Accept': "application/json"
+        }
+        disable_warnings(exceptions.InsecureRequestWarning)
+        resources = list()
+        response = requests.get(url,
```
viennaa: Would put this in a try except case, calls can go wrong in many ways tho

tools/Resources.py
```diff
@@ -0,0 +1,60 @@
...
+                                auth=HTTPBasicAuth(username=os.environ['USER'], password=os.environ['PASSWORD']),
+                                params=querystring,
+                                verify=False,
+                                headers=headers)
+        for resource in response.json()["resourceList"]:
```
viennaa: check with hasattr() if this key really exists before calling it

tools/Resources.py
```diff
@@ -0,0 +1,60 @@
...
+            res = dict()
+            res['name'] = resource["resourceKey"]["name"]
+            res['uuid'] = resource["identifier"]
```
viennaa: Same here, check if the keys do exist

tools/Resources.py
```diff
@@ -0,0 +1,60 @@
...
+            resources.append(res)
+        return resources
+
+    def get_project_id(self, target):
+        project_ids = list()
+        for project in self.get_vmfolders(target=target):
+            if project['name'].startswith('Project'):
```
viennaa: Check for the key

tests/TestCollectors.py
```diff
@@ -12,33 +12,52 @@
...
         metrics_yaml = YamlRead('tests/metrics.yaml').run()
         print(metrics_yaml)
-        #every collector got to be tested in here
-        random_prometheus_port = random.randrange(9000,9700,1)
+
+        # every collector got to be tested in here
+        random_prometheus_port = random.randrange(9000, 9700, 1)
         print("chosen testport: " + str(random_prometheus_port))
         for collector in metrics_yaml.keys():
+            print()  # nicer output
```
viennaa: remove, not in the master code anymore ;)

tests/TestCollectors.py
```diff
@@ -12,33 +12,52 @@
...
             print("\nTesting " + collector)
+
+            Vcenter.add_datacenter = MagicMock()
+            VropsCollector.create_resource_objects = MagicMock(return_value=Vcenter)
+            VropsCollector.get_adapter = MagicMock(return_value='{"name": "vcenter1", "uuid": "5628-9ba1-55e84701"}')
```
viennaa: Looks little bit confusing to me. Why would get_adapter give back vcenter1? Wouldn't that rather be Vcenter.uuid and Vcenter.name? 

What is an adapter here?

viennaa: Or should that be VropsCollector.get_vcenter rather than adapter?

resources/Vcenter.py
```diff
@@ -1,19 +1,16 @@
-from tools.get_resources import get_resources
-from resources.Cluster import Cluster
+from tools.Resources import Resources
```
viennaa: see cluster.py

tests/TestCollectors.py
```diff
@@ -12,33 +12,52 @@
...
+
+        # every collector got to be tested in here
+        random_prometheus_port = random.randrange(9000, 9700, 1)
         print("chosen testport: " + str(random_prometheus_port))
         for collector in metrics_yaml.keys():
+            print()  # nicer output
             print("\nTesting " + collector)
+
+            Vcenter.add_datacenter = MagicMock()
+            VropsCollector.create_resource_objects = MagicMock(return_value=Vcenter)
```
viennaa: Why do you need this? It blocks testing the whole functionality of create_resource_objects. Since you are mocking every single call in there, this should not be needed?

tests/TestCollectors.py
```diff
@@ -12,33 +12,52 @@
...
-            #start prometheus server to provide metrics later on
+            # test tool get_resources to create resource objects
+
+            Resources.get_datacenter = MagicMock(return_value={'name': 'datacenter1', 'uuid': '5628-9ba1-55e847050814'})
+            Resources.get_cluster = MagicMock(return_value={'name': 'cluster1', 'uuid': '3628-93a1-56e84634050814'})
+            Resources.get_hosts = MagicMock(return_value={'name': 'hostsystem1', 'uuid': '3628-93a1-56e84634050814'})
+            Resources.get_vmfolders = MagicMock(return_value={'name': 'vmfolder1', 'uuid': '3628-93a1-56e84634050814'})
+            Resources.get_virtualmachines = MagicMock(return_value={'name': 'vm1', 'uuid': '3628-93a1-56e8463404'})
+            Resources.get_resources = MagicMock(return_value={'name': 'vm1', 'uuid': '3628-93a1-56e8463404'})
+
```
viennaa: Cool! May you add more than one element into every mocked call? Would like to see the behavior tested with n+1 at least.

tests/TestCollectors.py
richardtief: In vrops, a vcenter is an adapter type. I kept the vmware syntax to avoid confusing those who are familiar with vmware. 
For my understanding `get_vcenter` is also the better name. I will change it.

viennaa: ok got it, maybe put in a comment that this is originally referring to adapters in vmware

richardtief: `get_adapter` returns a list of dicts like _[{'name': 'vcenter', 'uuid': '24543647' }, ...]_

the `create_resource_objects` makes from the get_requests objects. So you can call these like vcenter.name, vcenter.uuid

christopherhans: I would prefer Richards first initial approach with `get_adapter`, 'cause this is technically correct. I mean, in the end the whole purpose of our construct is, that no one has to take care of the exporter itself, if he wants to contribute a module(/metric). In this case no one has to dig into the `get_adapter` method.
vRops is capable of monitoring more than just vCenters. If this is needed at some point, we would add more **Adapters**. `get_adapter` would better fit in that case.
Just add as comment "In VMware environments a vCenter is an adapter." (Similar to what Tommy has written).

viennaa: Convinced, back to `get_adapter`

richardtief: ok, I roll back my changes

module/SampleCollector.py
```diff
@@ -1,8 +1,7 @@
 from BaseCollector import BaseCollector
-import requests
 import os
-from requests.auth import HTTPBasicAuth
 from prometheus_client.core import GaugeMetricFamily
+from VropsCollector import VropsCollector
```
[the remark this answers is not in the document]
richardtief: resolved

resources/Cluster.py
richardtief: resolved

resources/Datacenter.py
richardtief: resolved

resources/Host.py
richardtief: resolved

tools/Resources.py
richardtief: resolved

tools/Resources.py
richardtief: resolved

tools/Resources.py
richardtief: resolved

tools/Resources.py
richardtief: resolved

tools/Resources.py
richardtief: resolved

tests/TestCollectors.py
richardtief: resolved

tests/TestCollectors.py
richardtief: resolved

resources/Vcenter.py
richardtief: resolved

tests/TestCollectors.py
richardtief: resolved

tests/TestCollectors.py
richardtief: resolved

tools/Resources.py
```diff
@@ -0,0 +1,68 @@
...
+                                    verify=False,
+                                    headers=headers)
+            if hasattr(response.json(), "resourceList"):
+                for resource in response.json()["resourceList"]:
+                    res = dict()
+                    res['name'] = resource["resourceKey"]["name"]
+                    res['uuid'] = resource["identifier"]
+                    resources.append(res)
+            else:
+                raise AttributeError("There is no attribute: resourceList")
```
viennaa: Pretty cool :)

tools/Resources.py
```diff
@@ -0,0 +1,68 @@
...
+            'resourceKind': resourcekind,
+            'pageSize': '50000'
+        }
+        headers = {
+            'Content-Type': "application/json",
+            'Accept': "application/json"
+        }
+        resources = list()
+        try:
+            disable_warnings(exceptions.InsecureRequestWarning)
```
viennaa: move it out of try block

tools/Resources.py
viennaa: still missing hasattr

viennaa:
Thanks! 👍

PR #22, opened 2019-12-23
Test vrops exporter

richardtief:
start to write a test for checking the parameters at start


richardtief:
a further test to handle user errors ( bad options, missing arguments) is coming soon

richardtief:
https://docs.python.org/3/library/optparse.html#how-optparse-handles-errors

tests/TestLaunchExporter.py
```diff
@@ -0,0 +1,62 @@
...
+
+sys.path.append('.')
+from unittest import TestCase
+from vrops_exporter import parse_params
+
+
+class TestLaunchExporter(TestCase):
+
+    def test_with_cli_params_1(self):
+        sys.argv = ['prog', '-u', 'testuser', '-p', 'testpw31!', '-o', '1234']
```
viennaa: please add debug

tests/TestLaunchExporter.py
```diff
@@ -0,0 +1,62 @@
...
+        parse_params()
+        self.assertEqual(os.getenv('USER'), 'testuser', 'The user was not set correctly!')
+        self.assertEqual(os.getenv('PASSWORD'), 'testpw31!', 'The password was not set correctly!')
+        self.assertEqual(os.getenv('PORT'), '1234', 'The port was not set correctly!')
+
+    def test_with_cli_params_2(self):
+        sys.argv = ['prog', '--user', 'testuser', '--password', 'testpw31!', '--port', '1234']
```
viennaa: please add debug

tests/TestLaunchExporter.py
```diff
@@ -0,0 +1,62 @@
...
+        parse_params()
+        self.assertEqual(os.getenv('USER'), 'testuser', 'The user was not set correctly!')
+        self.assertEqual(os.getenv('PASSWORD'), 'testpw31!', 'The password was not set correctly!')
+        self.assertEqual(os.getenv('PORT'), '1234', 'The port was not set correctly!')
+
+    def test_with_cli_and_env_params(self):
```
viennaa: maybe slightly different naming:
cli params like: `cli_testuser`
env params like: `env_testuser` 
so they can be differentiated easier without specifically looking which one was defined where

tests/TestLaunchExporter.py
```diff
@@ -0,0 +1,62 @@
...
+        os.environ['PASSWORD'] = 'testps31!_2'
+        os.environ['PORT'] = '1123'
+        parse_params()
+        # cli params preferred
+        self.assertEqual(os.getenv('USER'), 'testuser', 'The user was not set correctly!')
+        self.assertEqual(os.getenv('PASSWORD'), 'testpw31!', 'The password was not set correctly!')
+        self.assertEqual(os.getenv('PORT'), '1234', 'The port was not set correctly!')
+
+        """
+    def test_with_wrong_options(self):
```
viennaa: I'd rename it: test_with_bogus_options
enable this one. The fail case is a success for this test.

vrops_exporter.py
```diff
@@ -30,13 +30,13 @@ def parse_params():
     if options.port:
         os.environ['PORT'] = options.port
 
-    if "PORT" not in os.environ and not options.port:
+    if "PORT" not in os.environ and not options.port or os.environ["PORT"] == ' ':
```
viennaa: is ` == '  '` and actual use case or what is the reason for this?

richardtief: I wonder if it could happen that the parameters are blank. I don't think so. I take this test case out.

tests/TestLaunchExporter.py
richardtief: test with debug option on

tests/TestLaunchExporter.py
richardtief: test with debug option off

viennaa: What do you mean by that? :D

richardtief: in this test I left out the option and checked that the option is not set.

tests/TestLaunchExporter.py
richardtief: in this test i set the option and checked that the option debug is set.

PR #23, opened 2020-01-03
HostSystemCollector

richardtief:


richardtief:
here ist one example for a HostSystem prometheus output: 
time for collecting:	3m 40.42s

`
HostSystemCollector{name="node004-bb095.cc.qa-de-1.cloud.sap",statkey="Memory|Usage (%)"} 2.448666572570801
HostSystemCollector{name="node004-bb095.cc.qa-de-1.cloud.sap",statkey="Memory|Balloon (KB)"} 0.0
HostSystemCollector{name="node004-bb095.cc.qa-de-1.cloud.sap",statkey="Memory|Compressed (KB)"} 0.0
HostSystemCollector{name="node004-bb095.cc.qa-de-1.cloud.sap",statkey="Memory|Guest Active (KB)"} 106298.3984375
HostSystemCollector{name="node004-bb095.cc.qa-de-1.cloud.sap",statkey="Memory|Consumed (KB)"} 6.5871045e+06
HostSystemCollector{name="node004-bb095.cc.qa-de-1.cloud.sap",statkey="Memory|Capacity Available to VMs (KB)"} 2.5082736e+08
HostSystemCollector{name="node004-bb095.cc.qa-de-1.cloud.sap",statkey="Memory|Total Capacity (KB)"} 2.68320016e+08
HostSystemCollector{name="node004-bb095.cc.qa-de-1.cloud.sap",statkey="Summary|Number of Running VMs"} 0.0
HostSystemCollector{name="node004-bb095.cc.qa-de-1.cloud.sap",statkey="Network|Packets Dropped"} 0.0
HostSystemCollector{name="node004-bb095.cc.qa-de-1.cloud.sap",statkey="Network|Transmitted Packets Dropped"} 0.0`

christopherhans:
Please stick to https://prometheus.io/docs/practices/naming/ for your metric naming.

richardtief:
For token based querys the time for collecting all resources in vrops + HostSystem metrics: 1m 56.73s

richardtief:
> Please stick to https://prometheus.io/docs/practices/naming/ for your metric naming.

will do.

module/HostSystemCollector.py
```diff
@@ -0,0 +1,62 @@
...
+                        {"name": "network_packets_dropped_rx_number",
+                         "statkey": "net|droppedRx_summation"},
+                        {"name": "network_packets_dropped_tx_number",
+                         "statkey": "net|droppedTx_summation"},
+                        {"name": "network_packets_dropped_percentage",
+                         "statkey": "net|droppedPct"}]
+
+    def collect(self):
+        metric_list = []
+        g = GaugeMetricFamily('HostSystemCollector', os.environ["TARGET"],
```
christopherhans: The naming scheme should something like `vrops_hostsystem_gauge`. 
If someone is searching for vrops metrics, he will start typing `vrops` and Prometheus will suggest all metrics from the vrops-exporter.

We **could** consider if we want to split the metric further. I.e.: `vrops_hostsystem_cpu_cores_total`, `vrops_hostsystem_cpu_sockets_info` and so on. While this is technically not necessary, it would simplify the searching process. @viennaa your opinion?

viennaa: @christopherhans i'm for splitting these metrics down as deep as needed. if we are in need of CPUs and sockes amounts then the metrics should be like you suggested. 

I also would not name the metric `_gauge`. For instrumenting them in alerts etc that information is not needed. To me it is also not clear, what this metric would mean, can you enlighten me?

christopherhans: `_gauge` would only be used in the first option, because it's the type of the metric and this metric would have multiple different metrics. But you are right, wouldn't be needed for an alert. 
Option 2 would drill this down anyway..

PR #28, opened 2020-02-06
Token based queries

richardtief:


InventoryBuilder.py
```diff
@@ -12,6 +12,7 @@
 
 class InventoryBuilder:
     def __init__(self, json):
+        self.token = self.get_token(self)
```
viennaa: please add flask entry to let token be queried

viennaa: and adapt the test case to use the token ;)

richardtief: I'm on my way

PR #31, opened 2020-02-07
implemented token based queries

richardtief:


tools/Resources.py
```diff
@@ -17,54 +15,52 @@ def get_resources(self, target, resourcekind, parentid):
...
                     res = dict()
                     res['name'] = resource["resourceKey"]["name"]
                     res['uuid'] = resource["identifier"]
                     resources.append(res)
-            else:
-                raise AttributeError("There is no attribute: resourceList")
+            except AttributeError as ar:
+                print("There is no attribute: resourceList", ar.args)
         except HTTPError as err:
             print("Request failed: ", err.args)
```
viennaa: Is there a specific reason, why you are not raising the error? If it fails it will return an empty list and the actual error will occur later.

InventoryBuilder.py
```diff
@@ -170,34 +172,57 @@ def create_resource_objects(self, vrops):
...
+                for resource in response.json()["adapterInstancesInfoDto"]:
+                    res = dict()
+                    res['name'] = resource["resourceKey"]["name"]
+                    res['uuid'] = resource["id"]
+                    res['adapterkind'] = resource["resourceKey"]["adapterKindKey"]
+                    adapters.append(res)
+            except AttributeError as ar:
+                print("There is no attribute adapterInstancesInfoDto " + str(ar.args))
         except HTTPError as err:
             print("Request failed: ", err.args)
```
viennaa: Same here, error is not being raised

PR #32, opened 2020-02-11
HostSystemCollector

richardtief:


richardtief:
- The `HostSystemCollector` is already registered in exporter.py

- metrics.yaml is up to date

- no further tests needed for `HostSystemCollector`

collectors/HostSystemCollector.py
```diff
@@ -0,0 +1,50 @@
+from BaseCollector import BaseCollector
+import os, time, json
+from pathlib import Path
+from prometheus_client.core import GaugeMetricFamily
+from tools.Resources import Resources
+
+
+class HostSystemCollector(BaseCollector, Resources):
```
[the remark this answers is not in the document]
richardtief: because of line 40. I have written the function get_metric into the resource class.

collectors/metrics.json
```diff
@@ -0,0 +1,65 @@
+{
```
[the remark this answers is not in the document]
richardtief: as good as done

tools/Resources.py
```diff
@@ -63,4 +63,23 @@ def get_virtualmachines(self, target, token, parentid):
...
+                                    verify=False,
+                                    headers=headers)
+            try:
+                for statkey in response.json()["values"][0]["stat-list"]["stat"]:
+                    if statkey["statKey"]["key"] is not None and statkey["statKey"]["key"] == key:
+                        return statkey["data"][0]
+            except AttributeError:
+                raise AttributeError("There is no attribute stat")
+        except HTTPError:
+            raise HTTPError("Request failed for statkey: " + key + " and target: " + target)
```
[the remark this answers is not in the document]
richardtief: that's the way I've handled it everywhere now. Then I have to think again about this error handling.

collectors/HostSystemCollector.py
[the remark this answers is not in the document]
viennaa: Please remove it from here. For L40 it is not needed, because it is explicitly called. I'd rather call Resources.get_hosts() instead of self, then this can be removed.

tools/Resources.py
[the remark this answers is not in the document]
viennaa: `except HTTPError as e:` and include str(e) in the error msg

InventoryBuilder.py
```diff
@@ -231,7 +244,7 @@ def get_token(self, target):
                                      headers=headers)
             try:
                 return response.json()["token"]
-            except AttributeError as ar:
-                print("There is no attribute token!", ar.args)
-        except (HTTPError, KeyError) as err:
-            print("Request failed: ", err.args)
+            except AttributeError:
+                raise AttributeError("There is no attribute token!")
```
viennaa: please check for return code. I was currently running into this error, which has completely being masked:
`{'message': 'Password Policy Violated. Reason: "Password expired: 0 day(s) ago. Password needs to be changed to proceed".', 'httpStatusCode': 400, 'apiErrorCode': 1513}`

In case of a none 200 return code I'd expect to see the whole information above.

Current code produces this: 
`TypeError: can only concatenate str (not "NoneType") to str` because there is no `token` in the response and therefore it is `None`.

collectors/HostSystemCollector.py
```diff
@@ -34,14 +34,12 @@ def collect(self):
         for hs in self.get_hosts():
             for statkey in self.statkeys:
                 r = Resources()
-                value = r.get_metric(self, target=self.target, token=self.token, uuid=self.hosts[hs]['uuid'],
-                                   key=statkey["statkey"])
+                value = r.get_value(self, target=self.target, token=self.token, uuid=self.hosts[hs]['uuid'],
+                                    key=statkey["statkey"])
+                if value is None:
+                    value = "0.0"
```
viennaa: 0 should be good for the label to set later, 0.0 is not needed

tools/Resources.py
```diff
@@ -63,7 +63,7 @@ def get_virtualmachines(self, target, token, parentid):
     def get_vmfolders(self, target, token):
         return self.get_resources(target, token, parentid=None, resourcekind="VMFolder")
 
-    def get_metric(self, target, token, uuid, key):
+    def get_value(self, target, token, uuid, key):
```
viennaa: how about `get_latest_stat`?

collectors/statkey.yaml
```diff
@@ -1,31 +1,30 @@
...
+      statkey: "mem|host_provisioned"
+    - label: "summary_running_VMs_number"
+      statkey: "summary|number_running_vms"
+    - label: "network_packets_dropped_rx_number"
+      statkey: "net|droppedRx_summation"
+    - label: "network_packets_dropped_tx_number"
+      statkey: "net|droppedTx_summation"
+    - label: "network_packets_dropped_percentage"
+      statkey: "net|droppedPct"
+
```
viennaa: that looks way better :). thanks!

viennaa: looks way better now :)

tools/Resources.py
richardtief: ok

tests/TestCollectors.py
```diff
@@ -53,7 +53,7 @@ def test_collector_metrics(self):
                                                       {'name': 'vm2', 'uuid': '5628-9ba1-55e847050814'}])
             Resources.get_resources = MagicMock(return_value=[{'name': 'resource1', 'uuid': '3628-93a1-56e8463404'},
                                                 {'name': 'resource2', 'uuid': '5628-9ba1-55e847050814'}])
-            Resources.get_value = MagicMock(return_value=1.0)
+            Resources.get_latest_stat = MagicMock(return_value=None)
```
viennaa: Shouldn't that be 1?

richardtief: yes sure, I had set to `None` for testing

viennaa: So it works either way?

PR #37, opened 2020-02-27
fetching statkey for multiple resources now

viennaa:
removed old flask entries for target and token
added target_tokens flask entry (gives a complete list now)
HostSystemCollector uses new method for fetching in parallel
key functionality is added in Resources

### DIFF
--- a/collectors/HostSystemCollector.py
+++ b/collectors/HostSystemCollector.py
@@ -21,17 +21,36 @@ class HostSystemCollector(BaseCollector):
 
         g = GaugeMetricFamily('vrops_hostsystem', 'testtext', labels=['datacenter', 'cluster', 'hostsystem', 'statkey'])
 
-        for hs in self.get_hosts():
-            target = self.hosts[hs]['target']
-            for statkey in self.statkey_yaml["HostSystemCollector"]:
-                r = Resources()
-                if os.environ['DEBUG'] >= '2':
-                    print("looking at " + str(self.hosts[hs]))
-                value = r.get_latest_stat(target=target, token=self.hosts[hs]['token'], uuid=self.hosts[hs]['uuid'], key=statkey["statkey"])
-                if value is None:
-                    value = "0"
-                if os.environ['DEBUG'] >= '1':
-                    print(self.hosts[hs]['name'], "--add statkey:", statkey["label"], str(value))
-                g.add_metric(labels=[self.hosts[hs]['datacenter'], self.hosts[hs]['parent_cluster_name'],
-                                     self.hosts[hs]['name'], statkey["label"]], value=value)
+        #make one big request per stat id with all resource id's in its belly
+        for target in self.get_hosts_by_target():
+            token = self.get_target_tokens()
+            token = token[target]
+            if not token:
+                print("skipping " + target + " in HostSystemCollector, no token")
+
+            uuids = self.target_hosts[target]
+            for statkey_pair in self.statkey_yaml["HostSystemCollector"]:
+                statkey_label = statkey_pair['label']
+                statkey = statkey_pair['statkey']
+                values = Resources.get_latest_stat_multiple(target, token, uuids, statkey)
+                if not values:
+                    print("skipping statkey " + str(statkey) + " in HostSystemCollector, no return")
+                    continue
+                for value_entry in values:
+                    #there is just one, because we are querying latest only
+                    metric_value = value_entry['stat-list']['stat'][0]['data'][0]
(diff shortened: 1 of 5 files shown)


collectors/HostSystemCollector.py
```diff
@@ -21,17 +21,36 @@ def collect(self):
...
+                print("skipping " + target + " in HostSystemCollector, no token")
+
+            uuids = self.target_hosts[target]
+            for statkey_pair in self.statkey_yaml["HostSystemCollector"]:
+                statkey_label = statkey_pair['label']
+                statkey = statkey_pair['statkey']
+                values = Resources.get_latest_stat_multiple(target, token, uuids, statkey)
+                if not values:
+                    print("skipping statkey " + str(statkey) + " in HostSystemCollector, no return")
+                    continue
```
richardtief: how can it happen that a statkey does not exist? it is more likely that a uuid of a resource cannot be found.

tools/Resources.py
```diff
@@ -63,6 +64,7 @@ def get_virtualmachines(self, target, token, parentid):
     def get_vmfolders(self, target, token):
         return self.get_resources(target, token, parentid=None, resourcekind="VMFolder")
 
+    #not recommended
     def get_latest_stat(self, target, token, uuid, key):
         url = "https://" + target + "/suite-api/api/resources/" + uuid + "/stats/latest"
         headers = {
```
richardtief: that can be removed. There is no need for it anymore.

tools/Resources.py
```diff
@@ -86,3 +88,32 @@ def get_latest_stat(self, target, token, uuid, key):
...
+                                     verify=False,
+                                     headers=headers)
+        except Exception as e:
+            print("Problem getting stats Error: " + str(e))
+            return False
+
+        if response.status_code == 200:
+            return response.json()['values']
+        else:
+            print("Return code not 200 for " + str(key) + ": " + response.json())
```
richardtief: Nice error handling. But we have to do it the same for all methods. I find the 200 as return code not clear enough as an error message. It would be better to know from which method the request just comes. We have several methods that make requests. What do you think?

tests/TestCollectors.py
```diff
@@ -55,6 +55,14 @@ def test_collector_metrics(self):
                                                 {'name': 'resource2', 'uuid': '5628-9ba1-55e847050814'}])
             Resources.get_latest_stat = MagicMock(return_value=1)
 
+            #mocking all values from yaml
+            statkey_yaml = YamlRead('collectors/statkey.yaml').run()
+            multiple_metrics_generated = list()
+            for statkey_pair in statkey_yaml["HostSystemCollector"]:
+                multiple_metrics_generated.append({ "resourceId": "3628-93a1-56e84634050814", "stat-list": { "stat": [{ "timestamps": [ 1582797716394 ], "statKey": { "key": statkey_pair['statkey'] }, "data": [ 88.0 ] }] } })
+                multiple_metrics_generated.append({ "resourceId": "5628-9ba1-55e847050814"  , "stat-list": { "stat": [{ "timestamps": [ 1582797716394 ], "statKey": { "key": statkey_pair['statkey'] }, "data": [ 44.0 ] }] } })
+            Resources.get_latest_stat_multiple = MagicMock(return_value=multiple_metrics_generated)
```
richardtief: nice!

collectors/HostSystemCollector.py
viennaa: Thought behind is to catch, if something goes wrong when issuing the API command. When an UUID is missing, it still will end up reporting 200, just tried it. If we would query for statkey, which does not exist, you would immediately see that. It wouldn't happen for tested code, but it will happen for others, specifying non-existent keys during delevopment.

tools/Resources.py
viennaa: I mean it is working and there for an easy single stat get. Not sure if we should throw it to the trash yet. However, I am going to remove `self` to make it callable.

tools/Resources.py
viennaa: What do you suppose instead? I don't want do have the whole traceback in here, the originating caller however would be interesting indeed.

collectors/HostSystemCollector.py
richardtief: O.K. Sure thing

tools/Resources.py
richardtief: Ok, that's right. But what does not work is to concatenate string and dict at `response.json()`.

PR #40, opened 2020-03-09
Properties for multiple resources

richardtief:
The test still fails due to this error:

```
Testing HostSystemStatsCollector
serving /vrops_list on 8000
serving /inventory on 8000
waiting for initial iteration
done: initial query
prometheus query port 9650
200 Server: BaseHTTP/0.6 Python/3.7.6
Date: Mon, 09 Mar 2020 08:05:29 GMT
Content-Type: text/plain; version=0.0.4; charset=utf-8



Testing HostSystemPropertiesCollector
waiting for initial iteration
done: initial query
prometheus query port 9651
500 Server: BaseHTTP/0.6 Python/3.7.6
Date: Mon, 09 Mar 2020 08:05:49 GMT
Connection: close
Content-Type: text/html;charset=utf-8
Content-Length: 467


----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 55860)
F.
======================================================================
FAIL: test_collector_metrics (__main__.TestCollectors)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/TestCollectors.py", line 122, in test_collector_metrics
    self.assertEqual(r.status, 200, "HTTP server return code should be 200")
AssertionError: 500 != 200 : HTTP server return code should be 200

----------------------------------------------------------------------
```


viennaa:
A little bit down the line this is the error: 

`ValueError: ("could not convert string to float: 'Powered On'", Metric(vrops_hostsystem_properties, testtest, gauge, `

richardtief:
> A little bit down the line this is the error:
> 
> `ValueError: ("could not convert string to float: 'Powered On'", Metric(vrops_hostsystem_properties, testtest, gauge, `

thanks!

tools/Resources.py
```diff
@@ -117,3 +117,163 @@ def get_latest_stat_multiple(target, token, uuids, key):
...
+        }
+        disable_warnings(exceptions.InsecureRequestWarning)
+        try:
+            response = requests.post(url,
+                                     data=json.dumps(payload),
+                                     verify=False,
+                                     headers=headers)
+        except Exception as e:
+            print("Problem getting property Error: " + str(e))
+            return False
```
viennaa: can we put this in an internal function to not duplicate this code? 
Could also be done for the stats collector.

collectors/property.yaml
```diff
@@ -0,0 +1,29 @@
+HostSystemPropertiesCollector:
```
viennaa: I really like the idea of having 3 types of different metrics!

collectors/HostSystemPropertiesCollector.py
```diff
@@ -0,0 +1,89 @@
...
+                        g.add_metric(
+                            labels=[self.hosts[host_id]['datacenter'], self.hosts[host_id]['parent_cluster_name'],
+                                    self.hosts[host_id]['name'], property_label],
+                            value=info_value)
+                    except ValueError:
+                        info = value_entry['data']
+                        info_value = 0
+                        g.add_metric(
+                            labels=[self.hosts[host_id]['datacenter'], self.hosts[host_id]['parent_cluster_name'],
+                                    self.hosts[host_id]['name'], property_label + ": " + info],
```
viennaa: info could be `'None'` if it is hitting the else condition in Resources. What would be the point of having this set to None?

collectors/HostSystemStatsCollector.py
```diff
@@ -17,24 +17,24 @@ def __init__(self):
 
     def collect(self):
         if os.environ['DEBUG'] >= '1':
-            print('HostSystemCollector ist start collecting metrics')
+            print('HostSystemStatsCollector ist start collecting metrics')
```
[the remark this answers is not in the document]
richardtief: I changed it to _HostSystemStatsCollector starts with collecting the metrics_.

tools/Resources.py
```diff
@@ -117,3 +117,163 @@ def get_latest_stat_multiple(target, token, uuids, key):
             print("Return code not 200 for " + str(key) + ": " + str(response.json()))
             return False
 
+    def get_latest_property(target, token, uuid, key):
+        url = "https://" + target + "/suite-api/api/resources/" + uuid + "properties"
```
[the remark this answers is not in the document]
richardtief: indeed

tools/Resources.py
```diff
@@ -117,3 +117,163 @@ def get_latest_stat_multiple(target, token, uuids, key):
             print("Return code not 200 for " + str(key) + ": " + str(response.json()))
             return False
 
+    def get_latest_property(target, token, uuid, key):
```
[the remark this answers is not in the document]
richardtief: Yes, there is no latest here. I name it `get_property`.

collectors/HostSystemPropertiesCollector.py
richardtief: prometheus can"t handle the value of `None`. I'm guessing it must be a number.

collectors/property.yaml
richardtief: me too

tools/Resources.py
richardtief: I've been thinking about that, too. I would like to handle this issue on a separate assignment. I would like to have only one class as interface to the vrops. That means, get token and get adapter will be included in this class. Then I have all methods in one class and can optimize duplicate code.

viennaa: Good point. Please create a task in jira for keeping this in mind.

PR #48, opened 2020-03-18
create collector_up metric

richardtief:


collectors/CollectorUp.py
```diff
@@ -0,0 +1,28 @@
+exp import os
```
viennaa: syntax

exporter.py
```diff
@@ -54,12 +55,41 @@ def parse_params():
 
 def run_prometheus_server(port, collectors, *args):
     start_http_server(int(port))
+    collector_list = list()
     for c in collectors:
         REGISTRY.register(c)
+        collector_list.append(REGISTRY._get_names(c))
```
viennaa: is there another way but using the internal function? How about using collectors and just skipping the CollectorUp one?

collectors/CollectorUp.py
```diff
@@ -0,0 +1,28 @@
...
+            self.collectors.append(collector[0])
+        self.metrics = metrics
+
+    def collect(self):
+        if os.environ['DEBUG'] >= '1':
+            print("Creating collector_up metrics")
+
+        g = GaugeMetricFamily('vrops_collector_up', 'testtext', labels=['collector'])
+
+        if not self.metrics:
```
viennaa: that's the `return False` from get_collector_up_information, nice!

exporter.py
```diff
@@ -54,12 +55,41 @@ def parse_params():
...
+
+    metrics = get_collector_up_information(port)
+    REGISTRY.register(CollectorUp(collector_list, metrics))
 
     while True:
         time.sleep(1)
 
+def get_collector_up_information(port):
```
viennaa: Couldn't that be in your collector? If you make sure, it gets registered after the other collectors that local connection would return the other metrics too.

exporter.py
richardtief: I check whether the collectors that are registered also return metrics. I have not found a better way to get names of timeseries the collector produces. But okay, I'll find another way.

exporter.py
richardtief: Yes indeed, I'm about to change it.

collectors/CollectorUp.py
richardtief: > What happens at the begin of the run? All collectors would first report 0, then 1, once they start delivering metrics? Shouldn't we give them a little time to avoid that?

No, the metrics are not generated until all collectors have run through once.

richardtief:
> What happens at the begin of the run? All collectors would first report 0, then 1, once they start delivering metrics? Shouldn't we give them a little time to avoid that?

No, the metrics are not generated until all collectors have run through once.

exporter.py
viennaa: Usually we should have the names, since we are having $collectors here already.

richardtief:
Every collector now has a self describing function that returns the metric name. I don't see any other way to get it at runtime without using internal functions. I compare registered or existing metric names with the output metrics of each collector. This ensures that each collector also provides metrics. I have tested it and it works properly.

exporter.py
```diff
@@ -66,9 +66,10 @@ def run_prometheus_server(port, collectors, *args):
     thread = Thread(target=InventoryBuilder, args=(options.atlas,))
     thread.start()
     collectors = [
-                HostSystemStatsCollector(),
                 SampleCollector(),
+                HostSystemStatsCollector(),
                 HostSystemPropertiesCollector(),
-                DatastoreStatsCollector()
+                DatastoreStatsCollector(),
+                CollectorUp()
```
viennaa: make a note, that new collectors have to be added _above_ `CollectorUp()`

richardtief:
I followed your suggestion and implemented it. It looks like this now:
```
# HELP vrops_collector_up testtext
# TYPE vrops_collector_up gauge
vrops_collector_up{collector="SampleCollector",metric="vrops_inventory_collection_iteration"} 1.0
vrops_collector_up{collector="HostSystemStatsCollector",metric="vrops_hostsystem_stats"} 1.0
vrops_collector_up{collector="HostSystemPropertiesCollector",metric="vrops_hostsystem_properties"} 1.0
vrops_collector_up{collector="HostSystemPropertiesCollector",metric="vrops_hostsystem_info"} 1.0
vrops_collector_up{collector="DatastoreStatsCollector",metric="vrops_datastore_stats"} 1.0
vrops_collector_up{collector="CollectorUp",metric="vrops_collector_up"} 1.0
```


Collectors can now yield multiple metrics.

PR #49, opened 2020-03-19
Atlas configfile handling fixes #30

### LINKED ISSUE #30: atlas configfile handling
@richardtief would you please add handling for the atlas configfile? 

* check if it has been provided
* check if it is a valid json
* check if it has the keys needed in InventoryBuilder

+ add Testcase to see if it fails and works like expected.

richardtief:


richardtief:
The test fails as expected because no netbox.json file is provided in /atlas/

viennaa:
2 Ways of solving this:
* we can put a dummy file (atlas.yaml) in / of this repository. The test can load it from ../netbox.json then.
* If you still want /atlas/netbox.json then use the DOCKERFILE and ADD to add respective file from repository root to the container.

PR #50, opened 2020-03-20
describe methods and wait function

viennaa:
adding describe function to avoid initial querying. gauges are now instantiated in init of collectors already.
removed code duplication for waiting stuff


collectors/DatastoreStatsCollector.py
```diff
@@ -14,14 +14,15 @@ def __init__(self):
             print("waiting for initial iteration")
         print("done: initial query")
         self.statkey_yaml = YamlRead('collectors/statkey.yaml').run()
```
richardtief: wait_for_inventory_data missing

PR #54, opened 2020-03-27
adding ClusterPropertyCollector

viennaa:


collectors/ClusterPropertiesCollector.py
```diff
@@ -0,0 +1,90 @@
...
+
+        for target in self.get_clusters_by_target():
+            token = self.get_target_tokens()
+            token = token[target]
+
+            if not token:
+                print("skipping " + target + " in " + self.name + ", no token")
+
+            uuids = self.target_clusters[target]
+            if 'number_metrics' in self.property_yaml[self.name]:
```
richardtief: good idea

richardtief:
The CollectorUp metric ist missing for the test. 

`vrops_collector_up{collector="ClusterPropertiesCollector",metric="vrops_cluster_properties"}`

collectors/ClusterPropertiesCollector.py
viennaa: figured it out when debugging some code with carl. if we don't have all metric types it would error then.

richardtief: We have to change it everywhere.

PR #55, opened 2020-03-30
add metrics for host_ramdisk_utilization

richardtief:


tests/TestCollectors.py
```diff
@@ -150,16 +150,17 @@ def test_collector_metrics(self):
...
 
             # check if there are more metrics being produced and they are not listed in metrics.yaml?!
             issubsetdifference = set(metrics).difference(metrics_yaml_list)
             self.assertTrue(set(metrics).issubset(metrics_yaml_list),
                             msg=collector + ": metric not covered by testcase, probably missing in yaml\n" + "\n".join(
                                 issubsetdifference))
+            # check if all metrics from yaml are here
+            supersetdifference = set(metrics_yaml_list).difference(metrics)
+            self.assertTrue(set(metrics).issuperset(metrics_yaml_list),
+                            msg=collector + ": missing metrics from yaml:\n" + "\n".join(supersetdifference))
```
viennaa: did you run into issues or why are you moving this?

tools/Resources.py
```diff
@@ -300,9 +300,9 @@ def get_latest_enum_properties_multiple(target, token, uuids, propkey, expected_
                 else:
                     latest_state = "unknown"
                 if latest_state == expected_state:
-                    d['data'] = 0
-                else:
                     d['data'] = 1
+                else:
+                    d['data'] = 0
```
viennaa: woops :D

tests/TestCollectors.py
richardtief: I have found for quite some time that asserting the missing metrics is more likely to fail first than asserting the added metrics in the `metrics.yaml.`

PR #61, opened 2020-04-03
better error handling

richardtief:
change order in register and for test

### DIFF
--- a/tools/Resources.py
+++ b/tools/Resources.py
@@ -254,6 +254,7 @@ class Resources:
                     else:
                         d['data'] = content[0]['data'][0]
                 else:
+                    # resources can go away, so None is returned
                     print("skipping resource for get", str(propkey))
 
                 properties_list.append(d)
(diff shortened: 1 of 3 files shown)


tools/Resources.py
```diff
@@ -247,14 +247,19 @@ def get_latest_number_properties_multiple(target, token, uuids, propkey):
...
-                    d['data'] = resource['property-contents']['property-content'][0]['values'][0]
+                content = resource['property-contents']['property-content']
+                if content:
+                    if 'values' in content[0]:
+                        d['data'] = content[0]['values'][0]
+                    else:
+                        d['data'] = content[0]['data'][0]
                 else:
-                    d['data'] = resource['property-contents']['property-content'][0]['data'][0]
+                    print("skipping resource for get", str(propkey))
```
viennaa, line 257: Please add comment, we are getting a `None` reply here and therefore need to skip. Most likely due to a deleted object or some connection problem?

PR #68, opened 2020-04-06
Make VROPS auth_source configureable

ThomasBuchinger:
Hi,
I am trying to use this exporter in our project, however we can't use the "Local" auth_source to connect to VROPS

I implemented an additional parameter to the `atlas.json` configuration file, to specify the auth_source (and also username and password). However I have no idea what an atlas-config is, so I am not sure if that is the correct place

### DIFF
--- a/InventoryBuilder.py
+++ b/InventoryBuilder.py
@@ -9,6 +9,7 @@ from tools.Resources import Resources
 import time
 import json
 import os
+import types
 
 
 class InventoryBuilder:
--- a/InventoryBuilder.py
+++ b/InventoryBuilder.py
@@ -14,6 +14,9 @@ import os
 class InventoryBuilder:
     def __init__(self, json, port):
         self.json = json
+        self.port = int(port)
+        self._user = os.environ["USER"]
+        self._password = os.environ["PASSWORD"]
         self.vcenter_dict = dict()
         self.target_tokens = dict()
         self.iterated_inventory = dict()
--- a/InventoryBuilder.py
+++ b/InventoryBuilder.py
@@ -37,7 +37,7 @@ class InventoryBuilder:
 
         @app.route('/vrops_list', methods=['GET'])
         def vrops_list():
-            return json.dumps(map(lambda x: x.server_name, self.vrops_list))
+            return json.dumps(list(map(lambda x: x.server_name, self.vrops_list)))
 
         print('serving /inventory on', str(self.port))
 
--- a/InventoryBuilder.py
+++ b/InventoryBuilder.py
@@ -14,9 +14,6 @@ import os
 class InventoryBuilder:
     def __init__(self, json, port):
         self.json = json
-        self.port = int(port)
-        self._user = os.environ["USER"]
-        self._password = os.environ["PASSWORD"]
         self.vcenter_dict = dict()
         self.target_tokens = dict()
         self.iterated_inventory = dict()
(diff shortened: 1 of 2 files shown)


InventoryBuilder.py
```diff
@@ -37,7 +35,7 @@ def run_rest_server(self):
 
         @app.route('/vrops_list', methods=['GET'])
         def vrops_list():
-            return json.dumps(self.vrops_list)
+            return json.dumps(map(lambda x: x.server_name, self.vrops_list))
```
ThomasBuchinger, line 38: should be converted to a list

ThomasBuchinger:
So atlas only provides the list of available VROPS. Any additional configuration should be in its own file. Got it. I think I have a good idea, how to do it.

ThomasBuchinger:
FYI: I am currently looking into implementing a config file like this one. 
```yaml
---
user: default_username
password: default_password
port: 8080
debug: true
profiles:
  - description: "Example vROPS profile"
    matches:
      - '.*' # match vrops-name with a regex
    user: vrops_username
    password: vrops_password
    auth_source: Local
    datacenter_type: Datacenter
    
    # This one is just an idea, if someone wants to limit which metrics
    # to collect it could be configured like this
    filter:
      HostSystem:
        include: [ 'site1-*' ]
        exclude: [ '.*-secret' ]
```

I guess sooner or later people want/need to be able to configure how InventoryBuilder works.  The basic idea it to configure everything in profiles and match that with the dynamic vrops-instance-list from atlas. I think the most flexible  way to match profiles to vrops-instances is a regex on the name (any other ideas?)

I am planning on one configuration file for both services (inventory and exporter). I think this would be easier to implement and use. Are you OK with one file for both or separate files?

viennaa:
Let's have the discussion in a separate issue.

tools/Resources.py
```diff
@@ -4,16 +4,23 @@
 
 class Resources:
     
-    def get_token(target):
+    def get_token(target, user=None, password=None, auth_source=None):
+        if auth_source == None:
```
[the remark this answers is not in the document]
viennaa: Please change this.

InventoryBuilder.py
```diff
@@ -129,7 +129,11 @@ def get_vrops(self):
         vrops_list = list()
         for target in netbox_json:
             if target['labels']['job'] == "vrops":
-                vrops = target['labels']['server_name']
+                vrops = types.SimpleNamespace()
+                vrops.server_name = target['labels']['server_name']
+                vrops.username    = target['labels'].get('username', None)
+                vrops.password    = target['labels'].get('password', None)
+                vrops.auth_source = target['labels'].get('auth_source', None)
                 vrops_list.append(vrops)
```
viennaa, line 137: Leveraging `target['labels']` to introduce username, pw and auth_source is not ideal. Since we are discussing #80 already I would propose the following:
Configure self.auth_source in InventoryBuilder globally and add the possibility to set it via ENV/CLI like we do everywhere, defaulting to `Local`.
That should enable you to use your auth source. Let's have the configfile / credential solution discussed first.

ThomasBuchinger: Yes. Once we have a config-file, we should move it there

viennaa: Just remove it and use the global var, if that wouldn't work for you, we can leave it in here for time being. Please add a `TODO: ..` here.

viennaa: Please decide for either one option.

viennaa:
Feel free to adapt to the rewritten code to get the feature in. Lots of things actually have changed. Closing, since there has been no activity for almost a year.

PR #74, opened 2020-04-14
catching JSONDecodeError

richardtief:
skip vm_id in VMPropertiesCollector

### DIFF
--- a/tools/Resources.py
+++ b/tools/Resources.py
@@ -249,25 +249,24 @@ class Resources:
                 if not response.json()['values']:
                     print("skipping propkey " + str(propkey) + ", no return")
                     return False
-                for resource in response.json()['values']:
-                    d = dict()
-                    d['resourceId'] = resource['resourceId']
-                    d['propkey'] = propkey
-                    content = resource['property-contents']['property-content']
-                    if content:
-                        if 'values' in content[0]:
-                            d['data'] = content[0]['values'][0]
-                        else:
-                            d['data'] = content[0]['data'][0]
-                    else:
-                        # resources can go away, so None is returned
-                        print("skipping resource for get", str(propkey))
-                    properties_list.append(d)
             except json.decoder.JSONDecodeError as e:
                 print("Catching JSONDecodeError for target:", str(target), "and key:", str(propkey),
                       "\nerror msg:", str(e))
                 return False
-
+            for resource in response.json()['values']:
+                d = dict()
+                d['resourceId'] = resource['resourceId']
+                d['propkey'] = propkey
+                content = resource['property-contents']['property-content']
+                if content:
+                    if 'values' in content[0]:
+                        d['data'] = content[0]['values'][0]
+                    else:
+                        d['data'] = content[0]['data'][0]
+                else:
+                    # resources can go away, so None is returned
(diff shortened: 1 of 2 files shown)


tools/Resources.py
```diff
@@ -88,7 +88,7 @@ def get_resources(self, target, token, resourcekind, parentid):
                                     headers=headers)
         except Exception as e:
             print("Problem connecting to " + target + "Error: " + str(e))
-            return False
+            return []
```
viennaa: That was long overdue :D

tools/Resources.py
```diff
@@ -180,7 +180,12 @@ def get_latest_stat_multiple(target, token, uuids, key):
             return False
 
         if response.status_code == 200:
-            return response.json()['values']
+            try:
+                return response.json()['values']
+            except json.decoder.JSONDecodeError as e:
+                print("Catching JSONDecodeError for target:", str(target), "and key:", str(key),
```
viennaa, line 186: please add spaces after the `:`

tools/Resources.py
```diff
@@ -240,24 +245,29 @@ def get_latest_number_properties_multiple(target, token, uuids, propkey):
...
-            for resource in response.json()['values']:
-                d = dict()
-                d['resourceId'] = resource['resourceId']
-                d['propkey'] = propkey
-                content = resource['property-contents']['property-content']
-                if content:
-                    if 'values' in content[0]:
-                        d['data'] = content[0]['values'][0]
+            try:
+                if not response.json()['values']:
```
viennaa, line 249: I'd say we only try catch this

tools/Resources.py
```diff
@@ -294,28 +304,32 @@ def get_latest_enum_properties_multiple(target, token, uuids, propkey, expected_
...
-                content = resource['property-contents']['property-content']
-                if content:
-                    if 'values' in content[0]:
-                        latest_state = content[0]['values'][0]
-                    else:
-                        latest_state = "unknown"
-                    if latest_state == expected_state:
-                        d['data'] = 1
+            try:
+                if not response.json()['values']:
```
viennaa, line 308: same here

tools/Resources.py
```diff
@@ -351,24 +365,28 @@ def get_latest_info_properties_multiple(target, token, uuids, propkey):
...
-            for resource in response.json()['values']:
-                d = dict()
-                d['resourceId'] = resource['resourceId']
-                d['propkey'] = propkey
-                content = resource['property-contents']['property-content']
-                if content:
-                    if 'values' in content[0]:
-                        info = content[0]['values'][0]
+            try:
+                if not response.json()['values']:
```
viennaa, line 369: and here

tools/Resources.py
richardtief: with a comma a space is inserted automatically

tools/Resources.py
richardtief: without catching JSONDecodeError and message?

viennaa: The problem always occurs here with `['values']`, so why take in the other stuff too?

richardtief: got it

PR #82, opened 2020-04-22
Add support for CustomDatacenters

ThomasBuchinger:
We use CustomDatacenters in out environment, which requires us to use a different query to get the datacenter.

I hardcoded the configuration option for now, because it would conflict with the other branch I am working on. Also I didn't find any existing unit tests for that part

### DIFF
--- a/InventoryBuilder.py
+++ b/InventoryBuilder.py
@@ -195,6 +195,8 @@ class InventoryBuilder:
             if os.environ['DEBUG'] >= '2':
                 print("Collecting vcenter: " + adapter['name'])
             vcenter = Vcenter(target=vrops, token=token, name=adapter['name'], uuid=adapter['uuid'])
+            # Note: if a cluster is in multiple Datacenters, parent_dc_name will be the last DC the cluster is a member
+            vcenter.add_custom_datacenter()
             vcenter.add_datacenter()
             for dc_object in vcenter.datacenter:
                 if os.environ['DEBUG'] >= '2':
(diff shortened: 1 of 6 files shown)


InventoryBuilder.py
```diff
@@ -183,11 +183,16 @@ def query_vrops(self, vrops):
...
+        datacenter_type = "Datacenter"
         for adapter in Resources.get_adapter(target=vrops, token=token):
             if os.environ['DEBUG'] >= '2':
                 print("Collecting vcenter: " + adapter['name'])
             vcenter = Vcenter(target=vrops, token=token, name=adapter['name'], uuid=adapter['uuid'])
-            vcenter.add_datacenter()
+            if datacenter_type == "Datacenter":
+                vcenter.add_datacenter()
+            elif datacenter_type == "CustomDatacenter":
+                vcenter.add_custom_datacenter()
```
viennaa, line 194: The `datacenter_type` is not needed. Just run a straight 

```
vcenter.add_datacenter()
vcenter.add_custom_datacenter()
```

If there is a resourcekind `CustomDatacenter` it will return, otherwise it won't.

resources/Vcenter.py
```diff
@@ -15,3 +15,8 @@ def add_datacenter(self):
         r = Resources()
         for dc in Resources.get_datacenter(r, target=self.target, token=self.token, parentid=self.uuid):
             self.datacenter.append(Datacenter(target=self.target, token=self.token, name=dc['name'], uuid=dc['uuid']))
+
+    def add_custom_datacenter(self):
+        r= Resources()
```
viennaa, line 20: please add space after `r`

InventoryBuilder.py
ThomasBuchinger: I tried to avoid duplicates in case the VROPS-User has access to normal Datacenters as well as CustomDatacenters. But you are right, it is not needed for the inventory, because the inventory ends up to be a dictionary with the UUID as the key.

One caveat though: In case we get the same cluster from a normal Datacenter and a CustomDatacenter, the `parent_dc_uuid` and `parent_dc_name` will be from the CustomDatacenter.
And parent_dc_name is used as a label in CluserStatsCollector

PR #94, opened 2020-05-12
Added VCenterPropertiesCollector & Updated VCenterStatsCollector

sunnyrarora:
Requesting a review -Thanks

### DIFF
--- a/collectors/VCenterPropertiesCollector.py
+++ b/collectors/VCenterPropertiesCollector.py
@@ -1,0 +1,64 @@
+from BaseCollector import BaseCollector
+import os
+from prometheus_client.core import GaugeMetricFamily, InfoMetricFamily
+from tools.Resources import Resources
+from tools.YamlRead import YamlRead
+from threading import Thread
+
+
+class VCenterPropertiesCollector(BaseCollector):
+    def __init__(self):
+        self.wait_for_inventory_data()
+        self.property_yaml = YamlRead('collectors/property.yaml').run()
+        self.name = self.__class__.__name__
+        # self.post_registered_collector(self.name, self.g.name, self.i.name + '_info')
+
+    def describe(self):
+        yield GaugeMetricFamily('vrops_vcenter_properties', 'testtest')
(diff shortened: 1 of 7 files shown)


collectors/VCenterPropertiesCollector.py
```diff
@@ -0,0 +1,64 @@
...
+    def do_metrics(self, target, g, i):
+        token = self.get_target_tokens()
+        token = token[target]
+        if not token:
+            print("skipping " + target + " in , no token")
+
+        for vc in self.get_vcenters():
+            uuid = self.vcenters[vc]['uuid']
+            target_vc = self.vcenters[vc]['name']
+            if 'info_metrics' in self.property_yaml[self.name]:
```
viennaa, line 49: Please add all type of metrics here:
* info
* number
* enum

The collector should be able to handle all our expected formats

collectors/VCenterStatsCollector.py
```diff
@@ -0,0 +1,60 @@
...
+                continue
+
+            try:
+                metric_value = int(values)
+                g.add_metric(labels=[target_vc, statkey_label], value=metric_value)
+
+            except (ValueError, TypeError):
+                info = values
+                metric_value = 0
+                g.add_metric(labels=[target_vc, statkey_label + ":" + info], value=metric_value)
```
viennaa, line 60: Why do you need a try catch here? Please add it straight.

collectors/VCenterPropertiesCollector.py
sunnyrarora: my concern here is - we are already using `summery|version` and `summary|vcfullname` in the collector as for - info_metrics.
For other two metrics like number and enum which Propkey i should pickup ? Only availability as in this picture. And also will that be required ? -Thanks
image

collectors/VCenterStatsCollector.py
sunnyrarora: Will work on this to resolve it. -Thanks

sunnyrarora:
Updated the requested changes and removed the try-catch block -Thanks

PR #97, opened 2020-05-26
1000 uuids stats chunk

viennaa:
if more than 1000uuids are queried at the same time, all but first 1000 are dropped. This is mitigating it and introducing threading for fetching stats, otherwise this will increase runtime significantly

### DIFF
--- a/tools/Resources.py
+++ b/tools/Resources.py
@@ -1,5 +1,7 @@
 from urllib3 import disable_warnings
 from urllib3 import exceptions
+from tools.helper import chunk_list
+from threading import Thread
 import requests
 import json
 import os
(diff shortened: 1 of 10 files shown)


tools/Resources.py
```diff
@@ -399,3 +364,65 @@ def get_latest_info_properties_multiple(target, token, uuids, propkey):
...
+        for uuid_list in uuids_chunked:
+            t = Thread(target=m.get_chunk, args=(uuid_list, url, headers, key, target))
+            thread_list.append(t)
+            t.start()
+        for t in thread_list:
+            t.join()
+        return m.return_list
+
+# helper class to allow to thread
+class MultipleThread0r:
```
richardtief, line 395: I'd like us to be more serious about this.

tools/helper.py
```diff
@@ -0,0 +1,3 @@
+def chunk_list(lst, n):
+    for i in range(0, len(lst), n):
+        yield lst[i:i + n]
```
richardtief, line 3: I think it make sense that we include the YamlRead.py also in helper.py. So that we have only one helper file.

tools/Resources.py
viennaa: :_(

tools/Resources.py
```diff
@@ -399,3 +364,65 @@ def get_latest_info_properties_multiple(target, token, uuids, propkey):
...
+        for uuid_list in uuids_chunked:
+            t = Thread(target=m.get_chunk, args=(uuid_list, url, headers, key, target))
+            thread_list.append(t)
+            t.start()
+        for t in thread_list:
+            t.join()
+        return m.return_list
+
+# helper class to allow to thread
+class ChunkFetcher:
```
richardtief, line 395: nice naming here!

PR #98, opened 2020-05-26
add config vm memory

richardtief:
there is no metric for configured vm memory on host or cluster level


viennaa:
can you please squash the commits?

PR #99, opened 2020-05-27
add metric for configured memory on vms

richardtief:


richardtief:
Whats wrong with the test coverage here? @viennaa

viennaa:
My assumption would be, that when you have created your branch the code coverage stuff was not in your master code. However it is a little bit weird, since the Token is setup in the repository itself.

PR #100, opened 2020-05-28
moved to queue

viennaa:
better debugging vor VMStats


viennaa:
@richardtief review pls

richardtief:
Why do we need an extra queue here? Is this because of the concurrency of requests?

viennaa:
The object was not as stateless as expected and ended up leaking memory. That's why I solved it with a queue.

PR #101, opened 2020-05-29
Semaphoring and queueing

viennaa:
kicked out ChunkFetcher since it introduced some memory leaking. Doing a classic threading way with a queue now.
Furthermore I've introduced Semaphores to steer parallel requests for VMStatsCollector and fetching resource methods. Speeding it up by roughly 40%.


viennaa:
Not ready for review yet, might be discontinued

richardtief:
looks promising

viennaa:
Discontinued. We have a lot of parallel requests already.

PR #110, opened 2020-07-08
added project ids to virtual machines

richardtief:


collectors/VMPropertiesCollector.py
```diff
@@ -42,6 +42,7 @@ def do_metrics(self, target, g, i):
         if not token:
             print("skipping", target, "in", self.name, ", no token")
         uuids = self.target_vms[target]
+        project_ids = Resources.get_project_ids(target, token, uuids)
```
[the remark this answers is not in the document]
viennaa: This part needs to be moved. For every iteration of do_metrics (as in every thread) this is being freshly fetched. If it would be fetched in collect() and passed to do_metrics() would only run once and not on a per metric base.

tests/TestCollectors.py
```diff
@@ -59,6 +59,8 @@ def test_collector_metrics(self):
                                                           {'name': 'resource2', 'uuid': '5628-9ba1-55e847050815'}])
         Resources.get_latest_stat = MagicMock(return_value=1)
         Resources.get_property = MagicMock(return_value="test_property")
+        Resources.get_project_ids = MagicMock(return_value=[{"3628-93a1-56e84634050814": "internal"},
```
viennaa, line 62: can we get a not internal test-case too?

collectors/VMPropertiesCollector.py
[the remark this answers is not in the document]
richardtief: Ok, thanks for your comments on that. The first thing I solved in the collector to not die at any line. The second is a bit tricky, because I need the newest vm_uuids at that point to ensure an up to date status of vms in projects. I guess the newest match between project ids and vms are crucial. Let me think about that some more.

tests/TestCollectors.py
richardtief: Thats not an issue.

collectors/VMPropertiesCollector.py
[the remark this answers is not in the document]
viennaa: You have to bear in mind, project ids aren't changing for VMs. This is a pretty static settings. Once a VM is deployed, it is hosted in the same project for its lifecycle in all.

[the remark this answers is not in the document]
richardtief: Thanks, that helps. With this information I can move things around.

collectors/VMPropertiesCollector.py
```diff
@@ -19,15 +19,17 @@ def describe(self):
...
-                             labels=['vccluster', 'datacenter', 'virtualmachine', 'hostsystem'])
+                             labels=['vccluster', 'datacenter', 'virtualmachine', 'hostsystem', 'project'])
         if os.environ['DEBUG'] >= '1':
             print(self.name, 'starts with collecting the metrics')
+        project_ids = self.get_project_ids_by_target()
 
         thread_list = list()
         for target in self.get_vms_by_target():
-            t = Thread(target=self.do_metrics, args=(target, g, i))
+            pids = project_ids[target]
```
viennaa, line 31: rename to `project_ids_target` (everywhere, where pid is used)
Usually I'd refer to a pid as a linux process id.

collectors/VMPropertiesCollector.py
```diff
@@ -55,11 +57,17 @@ def do_metrics(self, target, g, i):
                         continue
                     data = value_entry['data']
                     vm_id = value_entry['resourceId']
+                    project_id = "internal"
+                    if project_ids:
+                        for pid in project_ids:
```
viennaa, line 62: `pid` is not the correct term here. You are looping through `targets` since the data structure looks like `project_id[target][vm?]['vm_id']`
I'd rather expect a `for target in project_ids:`

collectors/VMPropertiesCollector.py
```diff
@@ -55,11 +57,17 @@ def do_metrics(self, target, g, i):
...
+                            if vm_id in pid.keys():
```
viennaa, line 63: `.keys()` is not needed

collectors/VMPropertiesCollector.py
richardtief: OK, fair enough.

collectors/VMPropertiesCollector.py
richardtief: No, the `for target` - loop is done in line 26 in `VMStatsCollector` respectively line 30 in `VMPropertiesCollector`. At this point, it is a list with `project ids` for every target. 

The data structure looks the following: `[{vm_id1: project_id1}, {vm_id2: project_id2}]` 

So we can talk about `for vm_id in project_ids:` That would make sense.

collectors/VMPropertiesCollector.py
richardtief: It works that way. So I wouldn't change it.

collectors/VMPropertiesCollector.py
viennaa: Sounds good, do it.

collectors/VMPropertiesCollector.py
viennaa: As long as you don't need the watch of the dynamic changes in the dict object there is no reason to use it. 
(Ref: https://docs.python.org/3.1/library/stdtypes.html#dict.keys)

viennaa:
Please have a look at the failing python test too:
https://github.com/sapcc/vrops-exporter/pull/110/checks?check_run_id=881066232

PR #113, opened 2020-07-27
add for-loop to cover all vms

richardtief:


tools/Resources.py
```diff
@@ -442,7 +442,8 @@ def get_project_id_chunk(q, uuid_list, url, headers, target, chunk_iteration):
             try:
                 for project in response.json()['resourcesRelations']:
                     p_ids = dict()
-                    p_ids[project["relatedResources"][0]] = project["resource"]["resourceKey"]["name"][
+                    for p in project["relatedResources"]:
+                        p_ids[p] = project["resource"]["resourceKey"]["name"][
```
viennaa, line 446: Good catch, awaiting the your test to cover the original problem, then we can merge.

viennaa:
Thanks for also adding the test case. Let’s ship it.

PR #114, opened 2020-07-27
Test for new targets in netbox fixes #30

### LINKED ISSUE #30: atlas configfile handling

richardtief:
The basic idea is to carry out a continuous test in the same time interval as the InventoryBuilder fetches resources, to ensure that all new targets have been included.


tests/TestNetboxTargets.py
```diff
@@ -0,0 +1,46 @@
...
+            self.assertEqual(r.status, 200, "HTTP server return code should be 200")
+            self.assertEqual(r.reason, "OK", "HTTP status should be OK")
+
+            data = r.read().decode()
+            self.assertEqual(json.dumps(vrops_list), data, "New targets where not taken into account!")
+            r.close()
+            c.close()
+            print("Targets:", data, "  ==  ", json.dumps(vrops_list))
+            print("Test for targets OK! \nRelaxing together with inventory...")
+            time.sleep(1810)
```
richardtief, line 46: At this point we could introduce a CLI option for setting the inventory sleep time at the beginning. How about that, @viennaa ?

richardtief:
The basic idea is to carry out a continuous test in the same time interval as the `InventoryBuilder` fetches resources, to ensure that all new targets have been included.

tests/TestNetboxTargets.py
viennaa: Yes, set the current value as default and if the option is specified, we set it. Please add the new option to the argument test too.

inventory.py
```diff
@@ -14,6 +14,7 @@ def parse_params():
     parser.add_option("-d", "--debug", help="enable debug", action="store_true", dest="debug", default=False)
     parser.add_option("-l", "--loopback", help="use 127.0.0.1 address instead of listen to 0.0.0.0",
                       action="store_true", dest="loopback")
+    parser.add_option("-s", "--sleep", help="specifiy sleep time for inventory builder", action="store", dest="sleep")
```
viennaa, line 17: would indicate its optional character, mentioning the default value

InventoryBuilder.py
```diff
@@ -131,12 +132,8 @@ def token():
     def get_vrops(self):
         with open(self.json) as json_file:
             netbox_json = json.load(json_file)
-        vrops_list = list()
-        for target in netbox_json:
-            if target['labels']['job'] == "vrops":
-                vrops = target['labels']['server_name']
-                vrops_list.append(vrops)
-        self.vrops_list = vrops_list
+        self.vrops_list = [target['labels']['server_name'] for target in netbox_json if
```
viennaa, line 135: Like the idea, any specific reason for this change?

richardtief: I did it this way on the test and I changed it right here. No special reason, just some clean code.

inventory.py
richardtief: Ok

InventoryBuilder.py
richardtief: And it is faster.

PR #115, opened 2020-07-29
move from static metric names

viennaa:
We want to have different metric names and remove the changing label.


collectors/HostSystemStatsCollector.py
```diff
@@ -1,53 +1,58 @@
...
+
     def __init__(self):
+        super().__init__()
+        self.vrops_entity_name = 'hostsystem'
         self.wait_for_inventory_data()
-        # self.post_registered_collector(self.__class__.__name__, self.g.name)
-
-    def describe(self):
-        yield GaugeMetricFamily('vrops_hostsystem_stats', 'testtext')
+        self.name = self.__class__.__name__
```
viennaa, line 13: I'd say we move this to __init__, then it is always there for sure

BaseCollector.py
```diff
@@ -162,3 +167,96 @@ def wait_for_inventory_data(self):
                 print("waiting for initial iteration: " + type(self).__name__)
         print("done: initial query " + type(self).__name__)
         return
+
+    def generate_gauges(self, metric_type, calling_class, vrops_entity_name, labelnames):
+        if not isinstance(labelnames, list):
+            print("Can't generate Gauges without label list, called from", calling_class)
+            return False
+        # switching between metric and property types
+        if metric_type == 'metric':
```
viennaa, line 176: Struggled about this naming a little bit. As we usually differentiating between `stats` and `property` I would expect `metric_type` checking  `stats`. The counterpart is checking `property`.

BaseCollector.py
```diff
@@ -162,3 +167,96 @@ def wait_for_inventory_data(self):
...
+                    gauges[property_suffix] = {
+                        'gauge': GaugeMetricFamily('vrops_' + vrops_entity_name + '_' + property_suffix.lower(),
+                                                   'vrops-exporter', labels=labelnames),
+                        'property': property_pair['property']
+                    }
+                return gauges
+            else:
+                return {}
+        else:
+            return False
```
viennaa, line 203: This is never evaluated in code. If it really hits this return value it would error in `do_metric` the first time the expected dict is looped. `gauges` then would be bool, instead of list. Maybe add some instance check `do_metrics` or check if it evaluates to something but False.

BaseCollector.py
```diff
@@ -4,9 +4,14 @@
 import os
 from tools.helper import yaml_read
 from tools.Resources import Resources
+from prometheus_client.core import GaugeMetricFamily, InfoMetricFamily, UnknownMetricFamily
+
 
 class BaseCollector(ABC):
 
+    def __init__(self):
+        self.vrops_entity_name = 'base'
```
viennaa, line 13: this does not seem to be used, any specific reason for this?

tools/Resources.py
```diff
@@ -310,14 +310,7 @@ def get_latest_enum_properties_multiple(target, token, uuids, propkey, expected_
...
                     if 'values' in content[0]:
-                        latest_state = content[0]['values'][0]
-                    else:
-                        latest_state = "unknown"
-                    if latest_state == expected_state:
-                        d['data'] = 1
-                    else:
-                        d['data'] = 0
-                    d['latest_state'] = latest_state
+                        d['value'] = content[0]['values'][0]
```
viennaa, line 313: I am glad this logic moved out of Resources 👍

collectors/HostSystemStatsCollector.py
richardtief: This is a old version. All collectors are now using the `describe()` method from `Basecollector`

BaseCollector.py
richardtief: Agreed.

BaseCollector.py
richardtief: This is used in `describe` method and will be overwritten in every collector.

BaseCollector.py
richardtief: Yes, the return is now also changed to a empty dict. This makes it more general usable in terms of adding new properties and statkeys to `collector_config`.

BaseCollector.py
```diff
@@ -200,12 +200,12 @@ def generate_gauges(self, metric_type, calling_class, vrops_entity_name, labelna
             else:
                 return {}
         else:
-            return False
```
viennaa: Introduce some error message, that we are hitting this condition. Isn't it missing for the `stats` section too?

You can also skip the two `else` and just give it a `return {}` at the end with some message, that given metric type is not handled. If one of the `if`s are matching, it won't reach this point.

tests/TestCollectors.py
```diff
@@ -153,13 +150,12 @@ def test_collector_metrics(self):
...
             c.request("GET", "/")
             r = c.getresponse()
-
             self.assertEqual(r.status, 200, "HTTP server return code should be 200")
             self.assertEqual(r.reason, "OK", "HTTP status should be OK")
 
             data = r.read().decode()
             data_array = data.split('\n')
-            metrics = list()
+            metrics = set()
```
viennaa, line 158: cool!

BaseCollector.py
richardtief: Done.

richardtief: Done.

viennaa:
Here we go, thanks for your contribution!

PR #118, opened 2020-08-20
update-architecture

richardtief:


viennaa (changes requested):
Thanks for taking the time to bring this up to date. Requesting the following changes:

Bullet point 1.
`Initially the inventory is built up and runs in a permanent update loop since then.

I would add `collect()` to SampleCollector too, since the collectors are implementing their own version.

PR #123, opened 2020-09-23
vc based exporter

richardtief:
The main idea is, to query the inventory with the right target. That means, starting multiple exporters with different targets and the exporters will fetch the related resources with the given target from inventory.


collectors/ClusterPropertiesCollector.py
```diff
@@ -94,3 +75,12 @@ def do_metrics(self, target, gauges, infos, states):
                     labels=[self.clusters[cluster_id]['name'],
                             self.clusters[cluster_id]['parent_dc_name'].lower()],
                     value={metric_suffix: info_value})
+
+        # self.post_metrics(self.g.name)
+        # self.post_metrics(self.i.name + '_info')
```
viennaa, line 80: We can remove this now

exporter.py
```diff
@@ -22,8 +22,10 @@ def parse_params():
                       dest="collectors")
     parser.add_option("-m", "--config", help="path to config to set default collectors, statkeys and properties for "
                                              "collectors", action="store", dest="config")
+    parser.add_option("-t", "--target", help="define target vcenter", action="store", dest="target")
```
viennaa, line 25: The way I got it this should be the target vrops, shouldn't it?

viennaa (changes requested):
Very solid implementation. Two minor things and we're good to go from my side.

collectors/ClusterPropertiesCollector.py
richardtief: With this, I removed all the `CollectorUp` Code.

exporter.py
richardtief: Your right. This have to be `vrops` instead of `vcenter`.

tools/vrops.py
```diff
@@ -7,7 +7,7 @@
 import os
 
 
-class Resources:
+class Vrops:
```
viennaa, line 10: Can we rename the file starting with uppercase `V`?

richardtief: Indeed.

tests/TestCollectorInit.py
```diff
@@ -12,7 +12,7 @@
 
 
 class TestCollectorInitialization(TestCase):
-    os.environ['TARGET'] = 'testhost.test'
+    os.environ.setdefault('TARGET', 'testhost.test')
```
viennaa, line 15: 👍  ;)

viennaa (approved):
Thanks, let's go!

PR #129, opened 2020-10-27
sharding VMStatsCollector

richardtief:
It turned out that the `VMStatsCollector` does not manage the scrapetime of 5 minutes. This is the case with a set of about 10000 or more virtual machines in a vCenter. 

The solution is to divide the `VMStatsCollector` into `CPU`, `memory`, `network`,  `virtualdisk` and `others` (including VMPropertiersCollector)


tests/TestCollectorInit.py
```diff
@@ -5,20 +5,20 @@
...
+import collectors.HostSystemStatsCollector
 from exporter import initialize_collector_by_name
 
 
 
 
 class TestCollectorInitialization(TestCase):
     os.environ.setdefault('TARGET', 'testhost.test')
-    collectors.VMStatsCollector.BaseCollector.get_target_tokens = MagicMock(return_value={'testhost.test': '2ed214d52'})
+    collectors.HostSystemStatsCollector.BaseCollector.get_target_tokens = MagicMock(return_value={'testhost.test': '2ed214d52'})
```
viennaa, line 16: Why is this changing?

tests/TestCollectors.py
```diff
@@ -37,8 +37,8 @@ def test_environment(self):
     def test_collector_metrics(self):
         metrics_yaml = yaml_read('tests/metrics.yaml')
         # every collector got to be tested in here
-        random_prometheus_port = random.randrange(9000, 9700, 1)
-        print("chosen testport: " + str(random_prometheus_port))
+        self.random_prometheus_port = random.randrange(9000, 9700, 1)
```
viennaa, line 40: what's the benefit of storing this in self?

tests/TestCollectors.py
```diff
@@ -191,9 +200,19 @@ def test_collector_metrics(self):
...
 
+        for collector in metrics_yaml.keys():
+            print("\nTesting " + collector)
+            if collector == 'VMStatsCollector':
+                for rubric in metrics_yaml[collector]['metrics']:
+                    os.environ['RUBRIC'] = rubric
+                    print("Testing rubric:", rubric)
+                    collector_testrun(rubric)
+            else:
+                collector_testrun()
```
viennaa, line 214: Please split the testing in two dedicated testable units
- with rubric
- without rubric

You have to define a really big inner scope function to accomplish doing everything in here and it will continue to be hard to read.

collectors/VMStatsCollector.py
```diff
@@ -8,12 +8,14 @@ class VMStatsCollector(BaseCollector):
     def __init__(self):
         super().__init__()
         self.vrops_entity_name = 'virtualmachine'
+        self.rubric = os.environ['RUBRIC']
         self.wait_for_inventory_data()
         self.name = self.__class__.__name__
 
     def collect(self):
         gauges = self.generate_gauges('stats', self.name, self.vrops_entity_name,
```
viennaa, line 16: please add hint, that rubric is needed for this particular collector only

collectors/VMStatsCollector.py
```diff
@@ -8,12 +8,14 @@ class VMStatsCollector(BaseCollector):
...
-                                      [self.vrops_entity_name, 'vcenter', 'datacenter', 'vccluster', 'hostsystem', 'project'])
+                                      [self.vrops_entity_name, 'vcenter', 'datacenter', 'vccluster', 'hostsystem',
+                                       'project'], rubric=self.rubric)
```
viennaa, line 18: What happens, if it is not provided for a collector, because you want all rubrics, not just a single one?

BaseCollector.py
```diff
@@ -113,21 +113,28 @@ def wait_for_inventory_data(self):
...
+            def iterate():
                 statkey_suffix = statkey_pair['metric_suffix']
                 gauges[statkey_suffix] = {
                     'gauge': GaugeMetricFamily('vrops_' + vrops_entity_name + '_' + statkey_suffix.lower(),
                                                'vrops-exporter', labels=labelnames),
                     'statkey': statkey_pair['statkey']
                 }
+            if calling_class == 'VMStatsCollector':
+                for statkey_pair in statkey_yaml[calling_class][rubric]:
+                    iterate()
```
viennaa, line 134: wouldn't it be better to define a method BaseCollector.generateGaugesByStatkey() and return gauges?

viennaa (changes requested):

tests/TestCollectorInit.py
richardtief: Just for testing purposes. Changed it back.

tests/TestCollectors.py
richardtief: To globally increase the prometheus testing port after each run. Otherwise this will lead to an error while unregistering metrics.

tests/TestCollectors.py
richardtief: I agree. But I wouldn't copy paste code just for testing this tiny difference. Let me remove the specific collector in there to make it a more general approach.

collectors/VMStatsCollector.py
richardtief: done

collectors/VMStatsCollector.py
richardtief: By default, `rubics=None`.  That means all metrics where considered. Look at the definition in `BaseCollector`.

BaseCollector.py
richardtief: I solved it by checking, if `rubrics=True`.

BaseCollector.py
```diff
@@ -16,6 +16,7 @@ def __init__(self):
             time.sleep(1800)
         self.target = os.environ['TARGET']
         # If metrics in collector-config are divided into rubrics
+        self.rubricated = False
```
viennaa, line 19: made my day

viennaa (approved):
With the default case configured in BaseCollector it is way better :).
Good to go from my side, let's see if we finally can get around our timeouts..

PR #135, opened 2020-12-15
keep exporter running

richardtief:
worked on errors which occurred in production

### DIFF
--- a/BaseCollector.py
+++ b/BaseCollector.py
@@ -82,9 +82,13 @@ class BaseCollector(ABC):
         return self.iteration
 
     def get_target_tokens(self):
-        request = requests.get(url="http://" + os.environ['INVENTORY'] + "/target_tokens")
-        self.target_tokens = request.json()
-        return self.target_tokens
+        try:
+            request = requests.get(url="http://" + os.environ['INVENTORY'] + "/target_tokens")
+            self.target_tokens = request.json()
+            return self.target_tokens
+        except requests.exceptions.ConnectionError as e:
+            logger.critical(f'No connection to inventory: {os.environ["INVENTORY"]} - Error: {e}')
+            return {}
 
     def get_clusters_by_target(self):
         cluster_dict = self.get_clusters(self.target)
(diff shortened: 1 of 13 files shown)


BaseCollector.py
```diff
@@ -107,11 +111,15 @@ def get_vms_by_target(self):
...
 
     def get_project_ids_by_target(self):
-        token = self.get_target_tokens()
-        token = token[self.target]
-        uuids = self.get_vms_by_target()
-        project_ids = Vrops.get_project_ids(self.target, token, uuids, self.collector)
-        return project_ids
+        try:
+            token = self.get_target_tokens()
+            token = token[self.target]
```
viennaa, line 116: Wouldn't this still fail, if `get_target_tokens` is returning `{}`?
Latest `get_project_ids` will fail because of the missing token and therefore the `ConnectionError` appears. As we now this is going to happen, avoid it.

exporter.py
```diff
@@ -105,10 +105,15 @@ def initialize_collector_by_name(class_name, logger):
...
-    return targets
+def get_targets(inventory):
+    try:
+        request = requests.get(url="http://" + os.environ['INVENTORY'] + "/vrops_list")
+        targets = request.json()
+        return targets
+    except requests.exceptions.ConnectionError as e:
+        logger.critical(f'No connection to {inventory} - Error: {e}')
+        logger.critical(f'Exit')
+        sys.exit(0)
```
viennaa, line 116: Just to get this right, are you doing this to enforce a pod restart?

viennaa:
I like the branch name 👍

BaseCollector.py
richardtief: No, it wouldn't fail. Because of line 17 `while` os.environ['TARGET'] not in self.get_target_tokens():`
The target will not be in the empty dictionary and thus will retry in 30 minutes. Or is restored earlier, as we have already observed.

exporter.py
richardtief: That is a good remark. What is the desired behaviour?

BaseCollector.py
richardtief: And to the second comment: No, every collector is now checking the token first. The code won't reach `get_project_ids`.

exporter.py
viennaa: Right now the pod would restart, which is not too bad, since it can't do anything anyway.
I would say let it retry 5 times with a 10seconds timeout and then exit.

exporter.py
```diff
@@ -106,14 +106,21 @@ def initialize_collector_by_name(class_name, logger):
...
+    while attempt == 1:
+        try:
+            request = requests.get(url="http://" + os.environ['INVENTORY'] + "/vrops_list")
+            targets = request.json()
+            return targets
+        except requests.exceptions.ConnectionError as e:
+            logger.critical(f'No connection to {inventory} - Error: {e}')
+            logger.critical(f'Trying again in 5sec.')
+            time.sleep(5)
+            attempt += 1
```
viennaa, line 120: Is it supposed to do one retry only?

richardtief: Ah, yes. It should be 5 times as you suggested.

viennaa (approved):

PR #137, opened 2021-01-26
added api response metric

richardtief:
Generating an api status code metric to check, if the vrops target is reachable to make further API calls. 

example out:
```
vrops_api_response{class="datastorepropertiescollector",collector="vrops-vc-a-0-default",target="vrops-vc-a-0.cc.qa-de-1.cloud.sap"} 200.0
vrops_api_response{class="vcenterstatscollector",collector="vrops-vc-a-0-default",target="vrops-vc-a-0.cc.qa-de-1.cloud.sap"} 200.0
vrops_api_response{class="hostsystemstatscollector",collector="vrops-vc-a-0-host,target="vrops-vc-a-0.cc.qa-de-1.cloud.sap"} 200.0
```

### DIFF
--- a/BaseCollector.py
+++ b/BaseCollector.py
@@ -132,10 +132,14 @@ class BaseCollector(ABC):
         return
 
     def create_http_response_metric(self, target, token, collector):
-        http_response_code = Vrops.get_http_response_code(target, token)
-        http_gauge = GaugeMetricFamily('vrops_http_response_code', 'vrops-exporter', labels=['target', 'class'])
-        http_gauge.add_metric(labels=[self.target, collector.lower()], value=http_response_code)
-        return http_response_code, http_gauge
+        api_responding = Vrops.get_http_response_code(target, token)
+        gauge = GaugeMetricFamily('vrops_api_response', 'vrops-exporter', labels=['target', 'class'])
+        gauge.add_metric(labels=[self.target, collector.lower()], value=api_responding)
+
+        if api_responding > 200:
+            logger.critical(f'API response {api_responding} [{collector}, {self.target}], no return')
+            return False, gauge
+        return True, gauge
 
     def generate_gauges(self, metric_type, calling_class, vrops_entity_name, labelnames, rubric=None):
         if not isinstance(labelnames, list):
--- a/BaseCollector.py
+++ b/BaseCollector.py
@@ -131,6 +131,12 @@ class BaseCollector(ABC):
         logger.info(f'-----Initial query done------: {self.collector}')
         return
 
+    def create_http_response_metric(self, target, token, collector):
+        http_response_code = Vrops.get_http_response_code(target, token)
+        http_gauge = GaugeMetricFamily('vrops_http_response_code', 'vrops-exporter', labels=['target', 'class'])
+        http_gauge.add_metric(labels=[self.target, collector.lower()], value=http_response_code)
+        return http_response_code, http_gauge
+
     def generate_gauges(self, metric_type, calling_class, vrops_entity_name, labelnames, rubric=None):
         if not isinstance(labelnames, list):
             logger.error(f'Cannot generate Gauges without label list, called from {calling_class}')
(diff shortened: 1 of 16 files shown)


collectors/ClusterStatsCollector.py
```diff
@@ -23,9 +23,14 @@ def collect(self):
...
 
+        http_response_code, http_gauge = self.create_http_response_metric(self.target, token, self.name)
+        logger.debug(f'HTTP response code in {self.name} for {self.target}: {http_response_code}')
+        yield http_gauge
+
         gauges = self.generate_gauges('stats', self.name, self.vrops_entity_name,
                                       ['vcenter', 'vccluster', 'datacenter'])
-        if not gauges:
+        if not gauges and http_response_code > 200:
+            logger.critical(f'HTTP response code in {self.name} for {self.target}: {http_response_code}, no return')
```
viennaa, line 33: Shouldn't we put it before `generate_gauges`?

collectors/ClusterPropertiesCollector.py
```diff
@@ -23,6 +23,14 @@ def collect(self):
             logger.warning(f'skipping {self.target} in {self.name}, no token')
             return
 
+        http_response_code, http_gauge = self.create_http_response_metric(self.target, token, self.name)
+        logger.debug(f'HTTP response code in {self.name} for {self.target}: {http_response_code}')
+        yield http_gauge
+
+        if http_response_code > 200:
+            logger.critical(f'HTTP response code in {self.name} for {self.target}: {http_response_code}, no return')
```
viennaa, line 31: I would move this log line and check for 200 to `create_http_response_metric()` to avoid duplicates. Same for `logger.debug`
Just return the gauge to yield it and a `True` or `False` variable which we use for continuing, like:

```
api_responding, http_gauge = self.create_http_response_metric(..)
if not api_responding:
    return
```

viennaa:
Remark: all tests are failing.

tools/Vrops.py
```diff
@@ -69,7 +92,7 @@ def get_adapter(target, token):
                 adapters.append(res)
         else:
             logger.error(f'Problem getting adapter {target} : {response.text}')
-            return False
+            return adapters
```
richardtief, line 95: fixed in this PR

viennaa: L84 should return adapters too.

BaseCollector.py
```diff
@@ -132,10 +132,14 @@ def wait_for_inventory_data(self):
...
     def create_http_response_metric(self, target, token, collector):
-        http_response_code = Vrops.get_http_response_code(target, token)
-        http_gauge = GaugeMetricFamily('vrops_http_response_code', 'vrops-exporter', labels=['target', 'class'])
-        http_gauge.add_metric(labels=[self.target, collector.lower()], value=http_response_code)
-        return http_response_code, http_gauge
+        api_responding = Vrops.get_http_response_code(target, token)
+        gauge = GaugeMetricFamily('vrops_api_response', 'vrops-exporter', labels=['target', 'class'])
+        gauge.add_metric(labels=[self.target, collector.lower()], value=api_responding)
+
+        if api_responding > 200:
```
viennaa, line 139: if `get_http_response_code` would run into the try except Exception, it would return 0, thus it won't match this condition. Shouldn't we check if there is a return value, if it is, return it, otherwise 500 (for the exception case)?

viennaa (commented):
Logging is better without duplication now. Some more comments in regards to the return values.

BaseCollector.py
richardtief: We could choose between `404 Not Found` or `408 Request Timeout`

viennaa: Why not 500? We are catching a random exception there. This is neither a timeout nor 404.

richardtief: The exporter application is the client in this scenario. So 4** HTTP codes are the right ones. If vrops not reachable:`Problem connecting to {target} - Error: {e}` I would return `408 Request Timeout`. `500` would mean, the server has an internal error. But we do not know, there is no response.

richardtief: But when I look at server HTTP codes, we could use `503 Service Unavailable`. I admit, this is maybe better.

viennaa: I'd take your 503 ;)

viennaa (approved):
Fire!

PR #138, opened 2021-02-09
only fetch receiving objects

viennaa:
objects need to be active objects, otherwise it could happen that vROps is fetching dated objects and then the exporter reports on these old dated metrics and objects too.


christopherhans (approved):
lgtm. Let's bring these numbers down.

PR #139, opened 2021-02-10
decrease retry for collectors

viennaa:
in case of a complete reset it could lead to a waiting collector, even though the inventory has finished the run. It just wasn't ready at the time the collector was respawned


richardtief (approved):
agree

PR #140, opened 2021-02-11
parallel calls for vrops targets

viennaa:
threading introduced for multiple runs to improve runtime.
shared timeout for threading
logging how long each target took to fetch its data

### DIFF
--- a/InventoryBuilder.py
+++ b/InventoryBuilder.py
@@ -12,10 +12,11 @@ logger = logging.getLogger('vrops-exporter')
 
 
 class InventoryBuilder:
-    def __init__(self, json, port, sleep):
+    def __init__(self, json, port, sleep, timeout):
         self.json = json
         self.port = int(port)
         self.sleep = sleep
+        self.timeout = int(timeout)
         self._user = os.environ["USER"]
         self._password = os.environ["PASSWORD"]
         self.vcenter_dict = dict()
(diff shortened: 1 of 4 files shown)


inventory.py
```diff
@@ -26,6 +26,8 @@ def parse_params(logger):
                       action="store_true", dest="loopback")
     parser.add_option("-s", "--sleep", help="specifiy sleep time for inventory builder, default: 1800", action="store",
                       dest="sleep")
+    parser.add_option("-t", "--timeout", help="specifiy timeout for fetching data from vROps, default: 600", action="store",
+                      dest="timeout")
```
richardtief, line 30: Is there a difference to `--sleep`?

InventoryBuilder.py
```diff
@@ -118,9 +119,35 @@ def query_inventory_permanent(self):
...
+                        if t[0] not in joined_threads:
+                            joined_threads.setdefault(t[1], round(time.time() - start_time))
+                if len(joined_threads.keys()) >= len(threads):
+                    break
+                time.sleep(1)
+                current_time = time.time()
+            else:
+                still_running = [t for t in threads if t[0].is_alive()]
+                for running_thread in still_running:
+                    logger.info(f"Timeout {timeout}s reached for fetching {running_thread[1]}")
```
richardtief, line 146: Alright, so use the timeout parameter to close the thread if it gets stuck forever?

richardtief (approved):
Overdue implementation, well implemented. I have a few questions about the code.

inventory.py
viennaa: it is. it defines how much time we give the inventory to fetch individual target vrops instances. If it is a bigger AZ, it takes longer for the inventory to fetch the data.
This is more like a failsafe, if something goes wrong in the thread, we kill it after $timeout seconds

PR #141, opened 2021-02-11
reduction of api calls and acceleration of the inventory building process

richardtief:
Reduce API calls to a minimum to keep the load on them as small as possible. At the same time, we need to update resources faster and more frequently so that dependent services work properly.

### DIFF
--- a/InventoryBuilder.py
+++ b/InventoryBuilder.py
@@ -149,29 +149,51 @@ class InventoryBuilder:
         logger.info(f'##############################################')
 
         vcenter = self.create_resource_objects(vrops, token)
+        if not vcenter:
+            return False
         self.vcenter_dict[vrops] = vcenter
         return True
 
-    def create_resource_objects(self, vrops, token):
-        for adapter in Vrops.get_adapter(target=vrops, token=token):
-            logger.debug(f'Collecting vcenter: {adapter["name"]}')
-            vcenter = Vcenter(target=vrops, token=token, name=adapter['name'], uuid=adapter['uuid'])
-            vcenter.add_datacenter()
-            for dc_object in vcenter.datacenter:
-                logger.debug(f'Collecting datacenter: {dc_object.name}')
-                dc_object.add_cluster()
-                for cl_object in dc_object.clusters:
-                    logger.debug(f'Collecting cluster: {cl_object.name}')
-                    cl_object.add_host()
-                    for hs_object in cl_object.hosts:
-                        logger.debug(f'Collecting host: {hs_object.name}')
-                        hs_object.add_datastore()
-                        for ds_object in hs_object.datastores:
-                            logger.debug(f'Collecting datastore: {ds_object.name}')
-                        hs_object.add_vm()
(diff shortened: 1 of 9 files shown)


InventoryBuilder.py
```diff
@@ -134,44 +161,65 @@ def query_inventory_permanent(self):
...
+    def create_resource_objects(self, target, token):
+        vrops = Vrops()
+
+        name, uuid = Vrops.get_adapter(target, token)
+        if not name:
+            return False
+        logger.debug(f'Collecting vcenter: {name}')
+
+        datacenter = Vrops.get_resources(vrops, target, token, [uuid], resourcekinds=["Datacenter"])
+        vccluster = Vrops.get_resources(vrops, target, token, [dc.get('uuid') for dc in datacenter],
```
viennaa, line 192: shouldn't we go with cluster here instead?

viennaa (commented):
I really like the approach, going to have a massive speed and api call impact. Just few comments.

viennaa:
Our tests are in an endless loop now.

InventoryBuilder.py
```diff
@@ -134,44 +161,65 @@ def query_inventory_permanent(self):
...
+                                        resourcekinds=["ClusterComputeResource"])
+        hosts = Vrops.get_resources(vrops, target, token, [cl.get('uuid') for cl in vccluster],
+                                    resourcekinds=["HostSystem"])
+        vms_and_ds = Vrops.get_resources(vrops, target, token, [hs.get('uuid') for hs in hosts],
+                                         resourcekinds=["Datastore", "VirtualMachine"])
+        vms = [vm for vm in vms_and_ds if vm.get('resourcekind') == "VirtualMachine"]
+        dss = [ds for ds in vms_and_ds if ds.get('resourcekind') == "Datastore"]
+
+        vcenter = Vcenter(target, token, uuid, name)
+        vcenter.add_datacenter(datacenter[0])
```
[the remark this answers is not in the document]
richardtief: We are sure. But I changed it to a loop anyway.

InventoryBuilder.py
```diff
@@ -199,8 +247,8 @@ def get_datacenters(self):
                     'parent_vcenter_uuid': vcenter.uuid,
                     'parent_vcenter_name': vcenter.name,
                     'vcenter': vcenter.name,
-                    'target': dc.target,
-                    'token': dc.token,
+                    'target': vcenter.target,
+                    'token': vcenter.token,
```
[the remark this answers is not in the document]
richardtief: To make API calls within the objects. No needed anymore.

InventoryBuilder.py
richardtief: Nope. We use the parent uuids to fetch the resources.

viennaa: i meant variable naming `vcccluster`

richardtief: suggestion pls

viennaa (approved):
Amazing finding, still. And shipping we do.

PR #144, opened 2021-02-17
http status message 

richardtief:


BaseCollector.py
```diff
@@ -86,6 +87,11 @@ def get_collection_times(self):
         self.collection_times = request.json()
         return self.collection_times
 
+    def get_inventory_api_response(self):
+        request = requests.get(url="http://" + os.environ['INVENTORY'] + "/api_response_code")
```
viennaa, line 91: This will get the response code of the last thread who just wrote something to `self.api_response`.
It maybe would make sense by target, but even then it is just the last request we are seeing here. Not sure if that makes sense.

resources/Datastore.py
```diff
@@ -19,4 +19,4 @@ def get_type(self):
         if "local" in self.name:
             return "local"
         else:
-            return "None"
+            return "Unknown"
```
viennaa, line 22: What is this about?

tools/Vrops.py
```diff
@@ -31,13 +31,13 @@ def get_token(target):
...
-            return False
+            return False, 522
 
         if response.status_code == 200:
-            return response.json()["token"]
+            return response.json()["token"], response.status_code
         else:
             logger.error(f'Problem getting token from {target} : {response.text}')
-            return False
+            return False, response.status_code
```
viennaa, line 40: What happens, if we don't have a status code?

tools/http_status_codes.py
```diff
@@ -0,0 +1,85 @@
...
+        503: ('Service Unavailable',
+              'The server cannot process the request due to a high load'),
+        504: ('Gateway Timeout',
+              'The gateway server did not receive a timely response'),
+        505: ('HTTP Version Not Supported', 'Cannot fulfill request.'),
+        506: (' Variant Also Negotiates',
+              'Transparent content negotiation for the request results in a circular reference.'),
+        510: ('Not Extended', 'Further extensions to the request are required for the server to fulfil it'),
+        511: ('Network Authentication Required', 'The client needs to authenticate to gain network access. '),
+        522: ('vROps not available', 'Problem connecting to vROps')
```
viennaa, line 84: mmyd :D

BaseCollector.py
```diff
@@ -138,8 +144,9 @@ def wait_for_inventory_data(self):
 
     def create_http_response_metric(self, target, token, collector):
         api_responding = Vrops.get_http_response_code(target, token)
-        gauge = GaugeMetricFamily('vrops_api_response', 'vrops-exporter', labels=['target', 'class'])
-        gauge.add_metric(labels=[self.target, collector.lower()], value=api_responding)
+        gauge = GaugeMetricFamily('vrops_api_response', 'vrops-exporter', labels=['target', 'class', 'message'])
```
viennaa, line 147:
```suggestion
        gauge = GaugeMetricFamily('vrops_api_response', 'vrops-exporter', labels=['target', 'class', 'http_message'])
```

BaseCollector.py
richardtief: You are right. Changed it to a dict similar to collection times.

resources/Datastore.py
richardtief: This is not related to this topic. But this annoys me since weeks. Datastore type `Unknown`  is much better understandable in this sense.

tools/Vrops.py
richardtief: Will simply not happen. The `response` object is there at this point. Otherwise `response.text `never resolves.

tools/http_status_codes.py
richardtief: I requested a own RFC for `522`.  😄

BaseCollector.py
```diff
@@ -144,7 +144,7 @@ def wait_for_inventory_data(self):
 
     def create_http_response_metric(self, target, token, collector):
         api_responding = Vrops.get_http_response_code(target, token)
-        gauge = GaugeMetricFamily('vrops_api_response', 'vrops-exporter', labels=['target', 'class', 'message'])
+        gauge = GaugeMetricFamily('vrops_api_response', 'vrops-exporter', labels=['target', 'class', 'http_message'])
```
viennaa, line 147: you've got to change the tests too now ;)

viennaa (approved):

PR #145, opened 2021-02-18
Revert "http status message "

richardtief:
Reverts sapcc/vrops-exporter#144


resources/Datastore.py
```diff
@@ -19,4 +19,4 @@ def get_type(self):
         if "local" in self.name:
             return "local"
         else:
-            return "Unknown"
+            return "other"
```
viennaa, line 22: Didn't you want this as `Unknown`?

richardtief: No. It was `None` before. Then I changed it to `Unknown`, which is not good also. Type = `other` is finally the better choice.

viennaa (approved):

PR #147, opened 2021-03-15
collectors 2.0

richardtief:
The goal was to get the API response metric from the real responses to ensure effective self-monitoring. This brought a cascade of changes to simplify metric generation:

1. There is only one method left for fetching latest values `get_latest_values_multiple()`. This method can now process multiple keys at once, which increases performance and decreases requests to vROps.
2. Completely rewritten `collect()` method that works generically for each stats or properties. Using `.get()` on dictionaries from fetched json data structures to avoid `KeyErrors` at runtime.
3. No rubrics for divided collectors anymore. Implemented now as inheritance of classes. 
4. Just one MetricFamily to simplify processing. This is accompanied by relabelling, which is implemented in the `collect()` method of the `PropertiesCollector`. 
5. Datastores in the inventory are now correctly assigned to Datacenter instead of HostSystems as before.
6. Resource objects are merged into one file `Resourcekinds.py`

Class Structure


collectors/PropertiesCollector.py
```diff
@@ -0,0 +1,95 @@
+from BaseCollector import BaseCollector
+import logging
+
+logger = logging.getLogger('vrops-exporter')
+
+
+class PropertiesCollector(BaseCollector):
+
+    def __init__(self):
+        super().__init__()
```
viennaa, line 10: is this really needed? It is inheriting `__init__` and will initialise ist. 
Question applies to all occurrences of the single super init.

BaseCollector.py
```diff
@@ -136,141 +143,36 @@ def wait_for_inventory_data(self):
...
+            return gauge
+        return gauge
+
+    def generate_metrics(self, label_names: list) -> dict:
+        collector_config = self.read_collector_config()
+        metrics = {m['key']: {'metric_suffix': m['metric_suffix'],
+                              'key': m['key'],
+                              'expected': m.setdefault('expected', None),
+                              'gauge': GaugeMetricFamily(
+                                  'vrops_' + self.vrops_entity_name + '_' + m['metric_suffix'].lower(),
```
viennaa, line 166: smart naming move

collectors/PropertiesCollector.py
```diff
@@ -0,0 +1,95 @@
...
+    def collect(self):
+        logger.info(f'{self.name} starts with collecting the metrics')
+
+        token = self.get_target_tokens()
+        token = token.setdefault(self.target, '')
+        if not token:
+            logger.warning(f'skipping {self.target} in {self.name}, no token')
+            return
+
+        uuids = self.get_uuids()
```
viennaa, line 27: shouldn't we rather use `get_project_ids` naming wise?

collectors/PropertiesCollector.py
```diff
@@ -0,0 +1,95 @@
...
+                labels = self.set_labels(resource_id, project_ids)
+
+                statkey = value_entry.get('statKey')
+                values_received.add(statkey)
+
+                metric_data = value_entry.get('data', [False])[0]
+                metric_value = value_entry.get('values', [False])[0]
+
+                if statkey in metrics:
+                    # enum metrics
```
viennaa, line 57: funny, our enums are still Gauges ;)

collectors/PropertiesCollector.py
```diff
@@ -0,0 +1,95 @@
...
+                    if metrics[statkey]['expected']:
+                        relabel = metrics[statkey]['gauge']._labelnames
+                        if 'state' not in relabel:
+                            relabel += ('state',)
```
viennaa, line 61: A little bit unlucky, that we have to do it here, but it is the tradeoff to have it in a central place.
Is the `,` here really needed `('state',)` ?

tools/Vrops.py
```diff
@@ -149,368 +124,120 @@ def get_resources(self, target, token, uuids, resourcekinds):
...
-        headers = {
-            'Content-Type': "application/json",
-            'Accept': "application/json",
-            'Authorization': "vRealizeOpsToken " + token
-        }
         payload = {
-            "resourceIds": uuids,
-            "propertyKeys": [propkey]
+            "resourceId": uuid_list,
+            "statKey": keys,
```
viennaa, line 196: The key is multiple keys, lmao

tools/Vrops.py
```diff
@@ -149,368 +124,120 @@ def get_resources(self, target, token, uuids, resourcekinds):
...
+            "PageSize": 500000
```
viennaa, line 197: is this the maximum? Or a random picked number?
How many entries are we getting back currently?

tools/Vrops.py
```diff
@@ -149,368 +124,120 @@ def get_resources(self, target, token, uuids, resourcekinds):
...
         for t in thread_list:
             t.join()
 
+        return_list = list()
+        response_status_code = 503
+
         while not q.empty():
-            project_ids += q.get()
+            returned_chunks = q.get()
+            response_status_code = returned_chunks[1] if returned_chunks[1] > 200 else 200
```
viennaa, line 176: can't we always set `response_status_code` to `returned_chunks[1]`?

tools/Vrops.py
```diff
@@ -149,368 +124,120 @@ def get_resources(self, target, token, uuids, resourcekinds):
...
+            return_list.extend(returned_chunks[0])
 
         logger.debug(f'Amount uuids: {len(uuids)}')
-        logger.debug(f'Fetched     : {len(project_ids)}')
+        logger.debug(f'Fetched     : {len({r.get("resourceId") for r in return_list})}')
```
viennaa, line 180: Are we still facing diffs here?

tests/collector_config.yaml
```diff
@@ -1,269 +1,280 @@
...
+    - 'VMStatsCPUCollector'
+    - 'VMStatsNetworkCollector'
+    - 'VMStatsVirtualDiskCollector'
+    - 'VMStatsMemoryCollector'
+    - 'VMStatsDefaultCollector'
+
+ClusterPropertiesCollector:
+  - metric_suffix: "configuration_dasConfig_admissionControlEnabled"
+    expected: "true"
+    key: "configuration|dasConfig|admissionControlEnabled"
```
viennaa, line 20: before rolling this out, we need to make sure to reflect the changes to the configmap in helm-charts

resources/Resourceskinds.py
```diff
@@ -0,0 +1,82 @@
...
+        else:
+            return "other"
+
+
+class Host:
+
+    def __init__(self, name, uuid):
+        self.uuid = uuid
+        self.name = name
+        self.vms = list()
```
viennaa, line 70: hehe sneaked in the datastore-un-mapping. 👍

resources/Resourceskinds.py
```diff
@@ -0,0 +1,82 @@
+
```
viennaa, line 1: Like the consolidation in one file.

tools/Vrops.py
```diff
@@ -149,368 +124,120 @@ def get_resources(self, target, token, uuids, resourcekinds):
...
 
-        url = "https://" + target + "/suite-api/api/resources/" + uuid + "/stats/latest"
-        headers = {
-            'Content-Type': "application/json",
-            'Accept': "application/json",
-            'Authorization': "vRealizeOpsToken " + token
-        }
-        disable_warnings(exceptions.InsecureRequestWarning)
+    def get_latest_stats_multiple(self, target: str, token: str, uuids: list, keys: list, collector: str):
+        return self.get_latest_values_multiple(target, token, uuids, keys, collector, kind='stats')
```
viennaa, line 189: it is really cool, that stats and properties are leveraging the same functionality now

tools/Vrops.py
```diff
@@ -149,368 +124,120 @@ def get_resources(self, target, token, uuids, resourcekinds):
...
+        logger.debug(f'collector: {collector}')
+        logger.debug(f'Amount keys : {len(keys)}')
+        for k in keys:
+            logger.debug(f'key  : {k}')
+
         for uuid_list in uuids_chunked:
             chunk_iteration += 1
-            t = Thread(target=Vrops.get_project_id_chunk,
-                       args=(q, uuid_list, url, headers, target, chunk_iteration))
+            t = Thread(target=Vrops._get_chunk,
```
viennaa, line 164: Am I getting this right, that `uuid_list` would be only one in case of properties? Hence no chunking

collectors/StatsCollector.py
```diff
@@ -0,0 +1,67 @@
...
+
+            for value_entry in resource.get('stat-list', {}).get('stat', []):
+                statkey = value_entry.get('statKey', {}).get('key')
+                values_received.add(statkey)
+
+                metric_data = value_entry.get('data', [None])[0]
+                if statkey in metrics and metric_data is not None:
+                    metrics[statkey]['gauge'].add_metric(labels=labels, value=metric_data)
+                else:
+                    no_match_in_config.add(statkey)
```
viennaa, line 56: What do you think of adding another metric, which could give us this information? It could only be created, if we are facing a "no match" and then we can at least easily visualise it in a dashboard. I still like the format here, but it will be buried in the logs. So either we have a kibana dashboard to visualise it as it is, or we will actually never see it.

viennaa (commented):
Wow! That I would name a core rewrite, 2.0 is also a pretty good fit. I am amazed by the code amount of code we are able to save here and how easy it actually was, to fetch multiple keys per POST payload instead of just one. 
It is way cleaner now and the new object concept stands out.

I only have a few questions and remarks, as this is already pretty well shaped. Well done and big thanks!

collectors/PropertiesCollector.py
richardtief: Yes, at a high level we want to have access to methods in the superclass. So we need `super()` here. Otherwise it will not work.

collectors/PropertiesCollector.py
richardtief: no, these are resource uuids, which are set to the actual resource uuids in the respective classes.

collectors/PropertiesCollector.py
richardtief: as they were before. We adjust this MetricFamily by ourselves.

collectors/PropertiesCollector.py
richardtief: Yes, it is a tuple object. And when you add one entry to a tuple, this is how you do it.

tools/Vrops.py
richardtief: Screen Shot 2021-03-18 at 15 37 06

tools/Vrops.py
richardtief: This is just a guess. I will do a test to ensure the right value here.

tools/Vrops.py
richardtief: Then the code will overwrite a 500 with a 200. This is not the desired behaviour. Currently, the code will return everything higher than 200 if a single chunk fail.

tools/Vrops.py
richardtief: It is a good indicator if resources are missing. I saw some small diffs during testing. But nothing to concern about.

tests/collector_config.yaml
richardtief: The deployment too. I prepared that already

tools/Vrops.py
richardtief: Yes, thats right. We are able to fetch 10000 resources at once and it is working. Stats need to be chunks. Don't know why.

collectors/StatsCollector.py
richardtief: Have you seen the output? I case of virtual disks and datastores there are many metrics which doesn't match. I also thought about this solution you suggest. Let me fix this in a follow up PR.

viennaa: For an initial check it is good, then we still would need to go through every region to check on it. at least, initially.

collectors/PropertiesCollector.py
viennaa: What you are doing here is an overwrite of `__init__` from the parent class. Then you are calling the `__init__` of the parent/super class and do nothing else. This call is needed, in case you want to enrich init with something. Unless this is done, you can remove `__ init__` here. 

VMPropertiesCollector however still needs it, since it is doing some enrichment, which wouldn't happen otherwise.

richardtief: Alright, got your point. Removed it from `StatsCollector` and `PropertiesCollector`.

collectors/StatsCollector.py
richardtief: I implemented a workaround to get the metrics. I'm not happy with it yet, but at least we're getting it for now.

tools/Vrops.py
```diff
@@ -149,368 +125,120 @@ def get_resources(self, target, token, uuids, resourcekinds):
...
-            "propertyKeys": [propkey]
-        }
         disable_warnings(exceptions.InsecureRequestWarning)
         try:
             response = requests.post(url,
                                      data=json.dumps(payload),
                                      verify=False,
-                                     headers=headers)
+                                     headers=headers,
+                                     timeout=30)
```
richardtief, line 211: Timeout had to be increased, otherwise vrops cannot deliver the large data structures on time.

collectors/StatsCollector.py
```diff
@@ -53,15 +50,16 @@ def collect(self):
                 if statkey in metrics and metric_data is not None:
                     metrics[statkey]['gauge'].add_metric(labels=labels, value=metric_data)
                 else:
-                    no_match_in_config.add(statkey)
-
-        if list(no_match_in_config):
-            logger.warning(f'Skipped keys, due to no match with config in {self.name}: {list(no_match_in_config)} '
-                           f'<-- compare with collector_config')
+                    new_metric_suffix = re.sub('[^a-zA-Z/s0-9\n.]', '_', statkey)
```
viennaa, line 53: Can we add a comment here? I am sure I forget why we are doing this. Or do some renaming, the actual `no_match_in_config` was self-explanatory.

viennaa (approved):
Let's ship it. Looking forward!

PR #148, opened 2021-03-22
fixes for collectors 2.0

richardtief:


collectors/InventoryCollector.py
```diff
@@ -13,6 +13,9 @@ def __init__(self):
         self.name = self.__class__.__name__
         self.wait_for_inventory_data()
 
+    def describe(self):
+        self.collect()
+
```
richardtief, line 18: This collector needs his own describe() method.

collectors/StatsCollector.py
```diff
@@ -37,30 +37,28 @@ def collect(self):
...
+        no_match_in_config = []
 
         for resource in values:
             resource_id = resource.get('resourceId')
             labels = self.set_labels(resource_id, project_ids)
 
             for value_entry in resource.get('stat-list', {}).get('stat', []):
                 statkey = value_entry.get('statKey', {}).get('key')
-                values_received.add(statkey)
+                norm_statkey = re.sub("[^a-zA-Z|_ ]+", "", statkey)
```
richardtief, line 48: Normalising keys here. Sometimes we get renamed metrics from API like `cpu:102|usage_average` instead of `cpu|usage_average`. This won't match the given statkey and producing way to much noisy metrics.

collectors/InventoryCollector.py
viennaa: Funny thing is, this isn't even be needed. If `describe` is missing, it will automatically call `collect()` at register time.

collectors/StatsCollector.py
viennaa: please add this as a comment

collectors/InventoryCollector.py
richardtief: In normal cases, yes, I know. But in this case the collector inherits the method of the BaseCollector. And this is not appropriate.

viennaa (approved):
Thanks for fixing the leftovers :)

PR #150, opened 2021-03-25
final touches on collectors 2.0

richardtief:
1. Added `return` on `describe()` in InventoryCollector. The collector does not want to work otherwise.
2. The PropertiesCollector should be able to process the non-matching keys too.
3. Resource-uuids should never be empty, otherwise the requests are failing. Very rare, but happens.

### DIFF
--- a/collectors/InventoryCollector.py
+++ b/collectors/InventoryCollector.py
@@ -11,15 +11,20 @@ class InventoryCollector(BaseCollector):
     def __init__(self):
         super().__init__()
         self.name = self.__class__.__name__
+        self.resourcekinds = ["vcenters", "datacenters", "clusters", "hosts", "datastores", "vms"]
 
     def describe(self):
-        return self.collect()
+        for resourcekind in self.resourcekinds:
+            yield GaugeMetricFamily(f'vrops_inventory_{resourcekind}', f'Amount of {resourcekind} in inventory')
+        yield CounterMetricFamily('vrops_inventory_iteration', 'vrops_inventory')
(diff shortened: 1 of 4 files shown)


collectors/InventoryCollector.py
```diff
@@ -11,10 +11,9 @@ class InventoryCollector(BaseCollector):
     def __init__(self):
         super().__init__()
         self.name = self.__class__.__name__
-        self.wait_for_inventory_data()
 
     def describe(self):
-        self.collect()
+        return self.collect()
```
viennaa, line 16: Funny this went in the code in the first place. No need to have `describe()` in here if we are calling `collect()`, that's done by prometheus implicitly if describe is missing.

collectors/PropertiesCollector.py
```diff
@@ -35,8 +39,8 @@ def collect(self):
             logger.warning(f'No values in the response for {self.name}. API code: {api_responding}')
             return
 
-        no_match_in_config = set()
         values_received = set()
+        no_match_in_config = []
```
viennaa, line 43:
```suggestion
        no_match_in_config = list()
```

Meanwhile I'd prefer this format :)

viennaa (commented):
Final thoughts on the final touches

collectors/InventoryCollector.py
viennaa: Please announce the 4 metrics we are using in `collect` here, to avoid initial calling of `collect()`

viennaa (approved):

PR #157, opened 2021-04-12
chunking vm requests

richardtief:
I have received feedback that virtual machines are missing. After a few tests I was able to confirm this and fix it. It looks like the amount of resource requests is limited to `10000`.

### DIFF
--- a/tools/Vrops.py
+++ b/tools/Vrops.py
@@ -142,7 +142,12 @@ class Vrops:
         return self.get_resources(target, token, parent_uuids, resourcekinds=["HostSystem"])
 
     def get_vms(self, target, token, parent_uuids):
-        return self.get_resources(target, token, parent_uuids, resourcekinds=["VirtualMachine"], data_receiving=True)
+        uuids_chunked = list(chunk_list(parent_uuids, 150))
+        vms = list()
+        for uuid_list in uuids_chunked:
+            vms.extend(self.get_resources(target, token, uuid_list, resourcekinds=["VirtualMachine"],
+                                          data_receiving=True))
+        return vms
 
     def get_latest_values_multiple(self, target: str, token: str, uuids: list, keys: list, collector: str,
                                    kind: str = None) -> (list, int, float):
--- a/tools/Vrops.py
+++ b/tools/Vrops.py
@@ -142,15 +142,16 @@ class Vrops:
         return self.get_resources(target, token, parent_uuids, resourcekinds=["HostSystem"])
 
     def get_vms(self, target, token, parent_uuids, vcenter_uuid):
-        amount_vms, api_responding, response_time = self.get_latest_stats_multiple(target, token, [vcenter_uuid],
-                                                                                   ['summary|total_number_vms'],
-                                                                                   'Inventory')
+        amount_vms, api_responding, _ = self.get_latest_stats_multiple(target, token, [vcenter_uuid],
+                                                                       ['summary|total_number_vms'],
+                                                                       'Inventory')
         number_of_vms = amount_vms[0].get('stat-list', {}).get('stat', [])[0].get('data', [0])[0] if \
             api_responding == 200 else 0
 
         # vrops cannot handle more than 10000 resource requests
-        if number_of_vms > 10000:
-            uuids_chunked = list(chunk_list(parent_uuids, int(round(len(parent_uuids) / 2, 0))))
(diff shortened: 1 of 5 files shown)


tools/Vrops.py
```diff
@@ -142,7 +142,12 @@ def get_hosts(self, target, token, parent_uuids):
         return self.get_resources(target, token, parent_uuids, resourcekinds=["HostSystem"])
 
     def get_vms(self, target, token, parent_uuids):
-        return self.get_resources(target, token, parent_uuids, resourcekinds=["VirtualMachine"], data_receiving=True)
+        uuids_chunked = list(chunk_list(parent_uuids, 150))
```
viennaa, line 145: Why 150? Shouldn't they be chunked at 1000?

richardtief: We decreasing the amount of parent uuids here. Hosts to be precise.

tools/Vrops.py
```diff
@@ -141,8 +141,25 @@ def get_cluster_and_datastores(self, target, token, parent_uuids):
     def get_hosts(self, target, token, parent_uuids):
         return self.get_resources(target, token, parent_uuids, resourcekinds=["HostSystem"])
 
-    def get_vms(self, target, token, parent_uuids):
-        return self.get_resources(target, token, parent_uuids, resourcekinds=["VirtualMachine"], data_receiving=True)
+    def get_vms(self, target, token, parent_uuids, vcenter_uuid):
+        amount_vms, api_responding, response_time = self.get_latest_stats_multiple(target, token, [vcenter_uuid],
```
viennaa, line 145:
```suggestion
        amount_vms, api_responding, _ = self.get_latest_stats_multiple(target, token, [vcenter_uuid],
```
```suggestion
        amount_vms, api_responding, response_time = self.get_latest_stats_multiple(target, token, [vcenter_uuid],
```

tools/Vrops.py
```diff
@@ -141,8 +141,25 @@ def get_cluster_and_datastores(self, target, token, parent_uuids):
...
+                                                                                   ['summary|total_number_vms'],
+                                                                                   'Inventory')
+        number_of_vms = amount_vms[0].get('stat-list', {}).get('stat', [])[0].get('data', [0])[0] if \
+            api_responding == 200 else 0
+
+        # vrops cannot handle more than 10000 resource requests
+        if number_of_vms > 10000:
+            uuids_chunked = list(chunk_list(parent_uuids, int(round(len(parent_uuids) / 2, 0))))
```
viennaa, line 153: So basically we are chunking every 2 hosts? Am I getting this right?

richardtief: We making two chunks with the half of the hosts each.

richardtief: We assuming a uniform distribution over the hosts, yes.

viennaa: Can we make it two tiers?

```
> 10000 -> 2
> 20000 -> 4
```

viennaa: or better even: every 10000 more we multiply chunks by two.

richardtief: Ok, you are right. I think also we exceeding 20000 VMs in the near future. I adjust this.

tools/Vrops.py
```diff
@@ -141,8 +141,27 @@ def get_cluster_and_datastores(self, target, token, parent_uuids):
...
-        return self.get_resources(target, token, parent_uuids, resourcekinds=["VirtualMachine"], data_receiving=True)
+    def get_vms(self, target, token, parent_uuids, vcenter_uuid):
+        amount_vms, api_responding, _ = self.get_latest_stats_multiple(target, token, [vcenter_uuid],
+                                                                       ['summary|total_number_vms'],
+                                                                       'Inventory')
+
+        number_of_vms = amount_vms[0].get('stat-list', {}).get('stat', [])[0].get('data', [0])[0] if \
+            api_responding == 200 else 0
+
+        # vrops cannot handle more than 10000 resource requests
```
viennaa, line 152:
```suggestion
        # vrops cannot handle more than 10000 uuids in a single request
```

viennaa (commented):
`dynamic split factor` mmyd 👍 

please check the inline comment

viennaa (approved):
Thanks for the implementation and finding this bug in the first place. Like it.

PR #159, opened 2021-05-12
operates even without nsxt

richardtief:
We ran into a problem when there were no nsxt resources in vrops.


christopherhans (approved):
lgtm

PR #162, opened 2021-06-03
current values only

richardtief:
Activated option that only current statistic values should be reported, i.e. the statistic values that have not published a value during the last collection cycles will be skipped. This option will be configurable in the future.

### DIFF
--- a/tools/Vrops.py
+++ b/tools/Vrops.py
@@ -248,13 +248,19 @@ class Vrops:
     def _get_chunk(q, uuid_list, url, headers, keys, target, kind, collector, chunk_iteration):
         logger.debug(f'chunk: {chunk_iteration}')
 
+        # Indicates whether to report only "current" stat values, i.e. skip the stat-s that haven't published any value
+        # during recent collection cycles.
+        current_only = True
+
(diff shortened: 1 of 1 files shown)


tools/Vrops.py
```diff
@@ -248,13 +248,19 @@ def get_latest_stats_multiple(self, target: str, token: str, uuids: list, keys:
     def _get_chunk(q, uuid_list, url, headers, keys, target, kind, collector, chunk_iteration):
         logger.debug(f'chunk: {chunk_iteration}')
 
+        # Indicates whether to report only "current" stat values, i.e. skip the stat-s that haven't published any value
+        # during recent collection cycles.
+        current_only = True
```
viennaa, line 253: Are you planning to make this configureable?

richardtief: Yes

viennaa (approved):
Thanks for testing and implementing. Don't forget, to add the new configurable option later on.

richardtief:
Ok. Noted here: https://github.com/sapcc/vrops-exporter/issues/163

PR #164, opened 2021-06-10
NSXTMgmtNodeCollector

richardtief:


InventoryBuilder.py
```diff
@@ -253,12 +258,18 @@ def create_nsxt_objects(self, target: str, token: str):
...
 
         nsxt_mgmt_cluster = Vrops.get_nsxt_mgmt_cluster(vrops, target, token, [a.uuid for a in nsxt_adapter])
+        nsxt_mgmt_nodes = Vrops.get_nsxt_mgmt_nodes(vrops, target, token, [c.uuid for c in nsxt_mgmt_cluster])
 
         for adapter_object in nsxt_mgmt_plane.adapter:
             for mgmt_cluster in nsxt_mgmt_cluster:
                 if mgmt_cluster.parent == adapter_object.uuid:
                     adapter_object.add_mgmt_cluster(mgmt_cluster)
                     logger.debug(f'Collecting NSX-T management cluster: {mgmt_cluster.name}')
+            for mgmt_cluster_object in adapter_object.management_cluster:
```
viennaa, line 268:
```suggestion
            for mgmt_cluster_object in nsxt_adapter.management_cluster:
```
This is already in code, but down in the get_nsxt_* functions you are using `nsxt_adapter` instead of `adapter_object`. Since it is the same I would find it easier to read with the exact same naming. Applies to all `adapter_object` occurences.

InventoryBuilder.py
```diff
@@ -175,6 +179,7 @@ def query_inventory_permanent(self):
             self.get_datastores()
             self.get_vms()
             self.get_nsxt_mgmt_cluster()
+            self.get_nsxt_mgmt_nodes()
```
viennaa, line 182: Maybe not for this PR, but something to discuss:

With growing complexity I find a bit confusing to have 
`Vrops.get_nsxt_mgmt_nodes
self.get_nsxt_mgmt_nodes`

Which pretty much looks the same, but ain't. Going with a standard approach here would be to rather use `self._get_nsxt_mgmt_nodes` instead.

What's your opinion? Also, if we do the big renaming bulk, it should go in a separate PR.

viennaa (changes requested):
Thanks for the implementation. Perfectly fits in the existing environment. 
One general remark and one regarding naming in particular.

InventoryBuilder.py
richardtief: Fair enough. Changed the naming for the object.

InventoryBuilder.py
richardtief: changed all functions from `get_...` to `expose_...`

collectors/StatsCollector.py
```diff
@@ -53,7 +53,7 @@ def collect(self):
             for value_entry in resource.get('stat-list', {}).get('stat', []):
                 statkey = value_entry.get('statKey', {}).get('key')
                 # Normalisation of keys retrieved from API (e.g. cpu:102|usage_average -> cpu|usage_average)
-                norm_statkey = re.sub("[^a-zA-Z|_ ]+", "", statkey)
+                norm_statkey = re.sub("[^a-zA-Z|_\- ]+", "", statkey)
```
viennaa: When does the `-` come into play?

InventoryBuilder.py
```diff
@@ -409,5 +420,28 @@ def get_nsxt_mgmt_cluster(self) -> dict:
...
+        return tree
+
+    def expose_nsxt_mgmt_nodes(self) -> dict:
+        tree = dict()
+        for nsxt_entry in self.nsxt_dict:
+            nsxt_mgmt_plane = self.nsxt_dict[nsxt_entry]
+            if not nsxt_mgmt_plane:
+                continue
+            tree[nsxt_mgmt_plane.target] = dict()
+            for nsxt_adapter in nsxt_mgmt_plane.adapter:
```
viennaa, line 433:
```suggestion
            for nsxt_adapter_object in nsxt_mgmt_plane.adapter:
```
Either you add `_object` here too, or you remove it in L263

collectors/StatsCollector.py
richardtief: `System Capacity|System-wide Logical Switch Ports|UsageCount`

InventoryBuilder.py
richardtief: outdated

viennaa (approved):
Good to go! Thanks.

PR #170, opened 2021-07-28
Adding internal vCenter uuids

viennaa:
to have unique VMs for balancing we need unique ids. It can happen, that same OpenStack UUID and name occurs in the vcenter and there is no way for automated tooling to distinguish, which one is the correct VM. Therfore introducing the internal VMWare uuids of the vcenter.


richardtief (approved):
Thanks. Our structure can easily deliver the values later, if necessary. Nice.

PR #172, opened 2021-07-29
Fixing errors and solving inconsistent labels

richardtief:
- Description of alerts added
- Secured the setting of labels
- Without alert definitions the `AlertCollector` does not run
- Removed the access to the protected attribute `_labelnames`
- Using the info metrics family to set alert labels dynamically

### DIFF
--- a/BaseCollector.py
+++ b/BaseCollector.py
@@ -6,7 +6,7 @@ import re
 import logging
 from tools.helper import yaml_read
 from tools.Vrops import Vrops
-from prometheus_client.core import GaugeMetricFamily
+from prometheus_client.core import GaugeMetricFamily, InfoMetricFamily
 
 logger = logging.getLogger('vrops-exporter')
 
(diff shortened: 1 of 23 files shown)


BaseCollector.py
```diff
@@ -249,12 +249,12 @@ def generate_metrics_enriched_by_api(self, no_match_in_config: list, label_names
...
         return gauges
 
-    def generate_alert_metrics(self, label_names: list) -> GaugeMetricFamily:
+    def generate_alert_metrics(self, label_names: list) -> InfoMetricFamily:
         if 'alert_name' not in label_names:
             label_names.extend(['alert_name', 'alert_level', 'status', 'alert_impact'])
-        gauge = GaugeMetricFamily(f'vrops_{self.vrops_entity_name}_alert', 'vrops-exporter',
-                                  labels=label_names)
-        return gauge
+        alert_metric = InfoMetricFamily(f'vrops_{self.vrops_entity_name}_alert', 'vrops-exporter',
```
viennaa, line 255: So we're completely moving to Info here? Can you elaborate? We would need to change `describe` then too, since it is still Gauge.

richardtief: This eliminates the need to access the protected `_labelnames` attribute. The `InfoMetricFamily` can accept prepared dicts with `label name` - `label value` pairs. With this we can easily set the individual alert labels. 

Describe needs to be changed, right.

collectors/AlertCollector.py
```diff
@@ -50,32 +55,26 @@ def collect(self):
...
-                alert_metric._labelnames += (label_name,)
-                labels.append(label_value)
-            alert_metric.add_metric(labels=labels, value=1)
-            alert_metric._labelnames = label_names
+            labels.extend([alert['alertDefinitionName'],
+                           alert['alertLevel'],
+                           alert['status'],
+                           alert["alertImpact"]])
+            alert_labels = self.generate_alert_label_values(alert)
+            alert_metric.add_metric(labels=labels, value=alert_labels)
```
richardtief, line 68: Solved quite simply.

viennaa: Just found it a minute ago. Totally makes sense then. Good catch!

viennaa (approved):
Nice improvement not using the internal variable anymore. Go ahead!

PR #173, opened 2021-08-24
inventoryBuilder memory leak fix

viennaa:
See commit message for further insights

### DIFF
--- a/InventoryBuilder.py
+++ b/InventoryBuilder.py
@@ -221,8 +221,8 @@ class InventoryBuilder:
         vcenter = self.create_vcenter_objects(vrops, target, token)
         nsxt_adapter = self.create_nsxt_objects(vrops, target, token)
 
-        self.vcenter_dict[vrops_short_name] = vcenter
-        self.nsxt_dict[vrops_short_name] = nsxt_adapter
+        self.vcenter_dict[target] = vcenter
+        self.nsxt_dict[target] = nsxt_adapter
 
         if iteration == 1:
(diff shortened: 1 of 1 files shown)


InventoryBuilder.py
```diff
@@ -221,8 +221,8 @@ def query_vrops(self, target, vrops_short_name, iteration):
         vcenter = self.create_vcenter_objects(vrops, target, token)
         nsxt_adapter = self.create_nsxt_objects(vrops, target, token)
 
-        self.vcenter_dict[vrops] = vcenter
-        self.nsxt_dict[vrops] = nsxt_adapter
+        self.vcenter_dict[vrops_short_name] = vcenter
+        self.nsxt_dict[vrops_short_name] = nsxt_adapter
```
richardtief, line 225: I changed all `vrops` naming in that function previously to `target`. I missed to do it for those two lines. 

Here is the commit: https://github.com/sapcc/vrops-exporter/commit/89edf43ff205c1cafadefc99d90149280abfb65f#



So I would suggest to name it `target` here too.

richardtief (changes requested):

InventoryBuilder.py
richardtief: And thanks for catching this bug. Appreciated.

viennaa: Going with target. Next time you can add a suggestion :)

richardtief (commented):
Thanks!

richardtief (approved):

PR #175, opened 2021-08-31
Expose internal names

viennaa:
For crossmapping in other tools it is helpful to expose the internal VMWare name of components. All components which have an internal name will now have the additional attribute .internal_name and the inventory will shout it out.

We need this for a cross-collaboration project to map the respective internal names to the real world. They will ask the inventory.


tools/Vrops.py
```diff
@@ -128,6 +128,13 @@ def get_resources(self, target: str, token: str,
                     resource_object.uuid = resource["resource"]["identifier"]
                     resource_object.resourcekind = resource["resource"]["resourceKey"]["resourceKindKey"]
                     resource_object.parent = resource.get("relatedResources", [None])[0]
+                    resource_identifiers = resource.get('resource').get('resourceKey').get('resourceIdentifiers')
```
richardtief, line 131: This might fail because the second and the third `.get()` is executed on a `None`. You should provide the alternative.

richardtief (changes requested):
Good catch. Didn't expect to find the info in the api. I have seen one small problem.

tools/Vrops.py
viennaa: Absolutely right, corrected.

richardtief (approved):
Thanks. Let's open the inventory for other services.

PR #176, opened 2021-09-01
exposing internal_name label for hosts

viennaa:
To simplify the whole setup it is way easier to expose the internal names and have them fetched. Only for hosts now, since it is not needed elsewhere yet.


richardtief (approved):
Agree. This label value is rather short, so it does not overload the metrics.

PR #182, opened 2021-10-26
Add timeout to all requests

max-len:
We recently observed exporters not returning the scrape request after 60 seconds, missing timeouts are most likely the reason for it. Adding standard timeout of 60 seconds, please adjust if necessary.

### DIFF
--- a/tools/Vrops.py
+++ b/tools/Vrops.py
@@ -31,6 +31,9 @@ class Vrops:
                                      verify=False,
                                      headers=headers,
                                      timeout=10)
+        except requests.exceptions.ReadTimeout as e:
+            logger.error(f'Request to {url} timed out. Error: {e}')
+            return False, 504
         except Exception as e:
             logger.error(f'Problem connecting to {target}. Error: {e}')
             return False, 503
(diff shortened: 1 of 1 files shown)


tools/Vrops.py
```diff
@@ -359,7 +374,11 @@ def get_project_ids(target: str, token: str, uuids: list, collector: str) -> (li
                                      data=json.dumps(payload),
                                      params=querystring,
                                      verify=False,
-                                     headers=headers)
+                                     headers=headers,
+                                     timeout=60)
```
viennaa, line 378: Question still persists, do we really need to wait `60s`?

max-len: What's @richardtief opionen about this?
It seems 60s was chosen for the longest running requests in `_get_chunk`, which I believe is a reasonable conservative value, i.e. between 'premature cutoff' and 'respond before scrape timeout'.
With adding it to all other requests we just make sure it doesn't get stuck, without potentially alter the current behaviour.

max-len: To get stats about the other requests we should export their `.elapsed.total_seconds()` and deduce the needed timeouts. That's a topic for a separate PR, though.

viennaa: When looking through the overall scrape times I can find `65s` in the bigger regions as a maximum for the entire scrape, not just one request.
As the requests here would sum up we are way over the `scrape_interval`, given every request takes the same time. I'd suggest to go with `30`.

max-len: Agreed

viennaa (approved):
Thanks! Good to go from my side.

PR #186, opened 2021-11-30
Support venv with shebang

max-len:


richardtief (approved):
Thanks for this input. This allows us the portability across different systems.

PR #189, opened 2021-12-15
consider changes in the inventory config

richardtief:
better naming for sddc and vcops objects

decrease overhead for InventoryCollector

### DIFF
--- a/collectors/InventoryCollector.py
+++ b/collectors/InventoryCollector.py
@@ -11,10 +11,9 @@ class InventoryCollector(BaseCollector):
     def __init__(self):
         super().__init__()
         self.name = self.__class__.__name__
-        self.resourcekinds = ["vcenters", "datacenters", "clusters", "hosts", "datastores", "vms"]
 
     def describe(self):
-        for resourcekind in self.resourcekinds:
+        for resourcekind in self.get_amount_resources()[self.target]:
             yield GaugeMetricFamily(f'vrops_inventory_{resourcekind}', f'Amount of {resourcekind} in inventory')
         yield CounterMetricFamily('vrops_inventory_iteration', 'vrops_inventory')
         yield GaugeMetricFamily('vrops_inventory_collection_time_seconds', 'vrops_inventory')
@@ -36,12 +35,9 @@ class InventoryCollector(BaseCollector):
 
     def amount_inventory_resources(self, target):
         gauges = list()
-        for resourcekind in self.resourcekinds:
+        for resourcekind, amount in self.get_amount_resources()[target].items():
             gauge = GaugeMetricFamily(f'vrops_inventory_{resourcekind}', f'Amount of {resourcekind} in inventory',
                                       labels=["target"])
-
-            type_method = getattr(BaseCollector, f'get_{resourcekind}')
-            amount = len(type_method(self, target))
             gauge.add_metric(labels=[target], value=amount)
             gauges.append(gauge)
         return gauges
--- a/tools/Vrops.py
+++ b/tools/Vrops.py
@@ -113,12 +113,10 @@ class Vrops:
         logger.debug(f'Getting {resourcekinds} from {target}')
         url = "https://" + target + "/suite-api/api/resources/bulk/relationships"
 
-        # If there is a config for the resourcekind in question, use it, otherwise give back the default.
-        q_spec = query_specs.get(resourcekinds[0]) if query_specs.get(resourcekinds[0]) else query_specs.get('default')
-        logger.debug(f'Using resource query specs: {q_spec}')
-        r_status_list = [rs for rs in q_spec.get('resourceStatus', [])]
-        r_health_list = [rh for rh in q_spec.get('resourceHealth', [])]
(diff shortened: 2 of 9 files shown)


collectors/InventoryCollector.py
```diff
@@ -36,12 +35,9 @@ def collect(self):
 
     def amount_inventory_resources(self, target):
         gauges = list()
-        for resourcekind in self.resourcekinds:
+        for resourcekind, amount in self.get_amount_resources()[target].items():
```
viennaa, line 38: Smart move to only calculate the length once and give it back.

viennaa (changes requested):
Looks good, like the change of only calculating the length once.

One change I'd like to see with the `query_specs` is, that InventoryBuilder should always use the default, unless specified different. I'd suggest to add another function which is doing this handling.

* Check `inventory_config` if there is a config for your `resourcekind` in question
* If there is, use it, otherwise give back the default. 
* put this in place wherever `inventory_config.get('query_specs', {})` is used

Currently `hostsystem` is defined separately, though it is actually using defaults. Any other `resourcekind` but `hostsystem` and `virtual_machines` would need a change of code, since only these two are considered in the config.

tools/Vrops.py
```diff
@@ -105,21 +105,24 @@ def get_resources(self, target: str,
...
                       ) -> (list, int):
         if not uuids:
             logger.debug(f'No parent resources for {resourcekinds} from {target}')
             return [], 200
-        logger.debug(f'Getting {resourcekinds} from {target}, query specs: {query_specs}')
+        logger.debug(f'Getting {resourcekinds} from {target}')
         url = "https://" + target + "/suite-api/api/resources/bulk/relationships"
+
+        # If there is a config for the resourcekind in question, use it, otherwise give back the default.
+        q_spec = query_specs.get(resourcekinds[0]) if query_specs.get(resourcekinds[0]) else query_specs.get('default')
```
viennaa, line 117: What happens if multiple resourcekinds are getting passed? Wouldn't we just look at index 0?

richardtief: Only one query specification can be used per query, even if there are several different ones for the resourcekinds. So the query specifications of the first mentioned resourcekind are always used.

viennaa: Wouldn't that mean, that the config possibly wouldn't fit for the other resourcekinds if multiple are passed?

richardtief: Yes. The only thing we can do is to make only one request per resourcekind at a time. But this again increases the overhead on the API.

richardtief: But I also welcome your suggestion on this.

viennaa: Well if we have different settings specified in `inventory_config` we must not ignore it and take the first one only. That would break every expected behaviour.

We should sort it out here, when we have multiple resourcekinds defined in `inventory_config` we've got to make single calls. The rest should be grouped und queried with the default query specs.

Make single resourcekind calls for the resourcekinds defined explicitly in `inventory_config` and for the rest we do a default call. To avoid duplication here we should move the actual `payload`-`request` handling to a dedicated function. like `get_resources_request` and feed in the seperated query specs, resourcekinds which are fitting (multiple for default, single resourcekinds for specifics).

richardtief:
I used the wrapper functions to configure `get_resources` for the resource kind in question.

tools/Vrops.py
```diff
@@ -187,17 +185,23 @@ def get_resources(self, target: str,
             return resources, response.status_code
 
     def get_datacenter(self, target, token, parent_uuids, query_specs):
-        return self.get_resources(target, token, adapterkind="VMWARE", resourcekinds=["Datacenter"],
-                                  uuids=parent_uuids, query_specs=query_specs)
+        resourcekind = 'Datacenter'
+        q_specs = query_specs.get(resourcekind) if resourcekind in query_specs else query_specs.get('default', {})
+        return self.get_resources(target, token, adapterkind="VMWARE", resourcekinds=[resourcekind],
+                                  uuids=parent_uuids, query_specs=q_specs)
```
viennaa, line 191: I like this behaviour more now.
We are introducing some code duplication here though.

Why aren't we doing this in a dedicated function and are calling this function at the beginning of `get_resources`?

Pretty much the same behaviour, we're passing in `resourcekinds` and are giving back the `q_specs`. We then can handle it centrally.

viennaa (changes requested):
Thanks for picking this up after a while. Let's discuss, if we can't avoid the duplication here.

tools/Vrops.py
richardtief: I know. This is not as DRY as I would like it to be. But there is a tradeoff. 

Inside `get_resources` is not the best way. The complexity starts with a list of `resourcekinds` like `sddc` and `vcops`.  They are always requested with default specifications. But we add more complexity to this function when we check each `resourcekind` inside the function. Let us do it outside the function. I have proposed a new commit.

viennaa: That's the spirit, like it :)

viennaa (approved):
Good to go, let's ship it together with #190

PR #190, opened 2022-02-04
Adding VM instance_uuid label

viennaa:
* instance_uuid label reflects the internal OpenStack uuid of the VM. It
is unique across multiple vCenters and therefore can be used to uniquely
detect a VM. Current virtualmachine label does not guarantee this across
vMotions.
* incoorporated slight improvements to get_resources. There were couple
of not needed .get() functions, whereas a KeyError would have taken
place way before already. The failure case also wasn't catched in case
get would have returned null


richardtief (approved):
Nice enhancement. Thanks.

PR #192, opened 2022-02-08
using defaultdicts

richardtief:
Prevent dicts from being emptied


viennaa (approved):
Good catch and solution. Thanks.

PR #194, opened 2022-02-14
README updated

richardtief:


richardtief:
@viennaa @christopherhans Please read over once, if there are any errors in it.

README.md
```diff
@@ -2,43 +2,142 @@
...
+/api_response_codes                               # HTTP response codes per resourcekind GET request
+/target_tokens                                    # dict with vrops: auth token
+```
+#### exporter
+
+The second component are the collectors that exist for each resourcekind as well as for metrics, properties and alerts. Each collector is responsible for only one task per resourcekind and one type of values. First, the resourcekinds in question are queried at the inventory's internal API. In the second step, the desired values, properties or alerts are queried. Each collector performs only one task - one resourcekind and one type from the three different values. First, the resourcekinds in question are queried at the inventory's internal API. In the second step, the values, properties or alarms are queried. From these, the Prometheus metrics are generated. To complete the picture, the metrics are enriched with the labels from the resourcekind relationships created in the inventory.
+
+![](images/architecture.png)
+
+To avoid multiple implementations of functionality, the collectors follow an inheritance structure.
```
viennaa, line 102: Path to `/atlas/netbox.json` is specific to us.

Actually just a netbox-like file is needed there. I would put in a file as input for `inventory.py` and an arrow from InventoryBuilder to that file with `<<use>>`. Reference to documentation for internal format.
Maybe some dotted line around the atlas file to indicate, that we are providing this with a service.

README.md
```diff
@@ -2,43 +2,142 @@
...
+
+![](images/collectors.png)
```
viennaa, line 104: Nice!

README.md
```diff
@@ -2,43 +2,142 @@
...
+
+## Running the software
```
viennaa, line 106: Should we move this to the top? If not, we should give it a table of contents at the beginning of this documentation.

README.md
```diff
@@ -2,43 +2,142 @@
...
+
+#### **inventory**
+
+The inventory retrieves all VMware resources by `name` and `uuid` from the vROps application interface, preserving the original parent-child relationships.         Likewise, it reads the atlas config file to know all the DNS names of the vrops targets.
```
viennaa, line 110: Isn't this duplicate information to the inventory section above?

README.md
```diff
@@ -103,21 +219,17 @@ The exporter consists of two main components:
 
 For running this in kubernetes (like we do), you might want to have a look at our [helm chart](https://github.com/sapcc/helm-charts/tree/master/prometheus-exporters/vrops-exporter)
```
viennaa, line 220: This should go to a dedicated section:
```Run in Kubernetes```

viennaa (changes requested):
Thanks for tackling this long due item. I've added my remarks.

PR #196, opened 2022-02-15
target label added for nsxt collectors

richardtief:
NSX-T adapters do not inherently show the vROps in which they are added. To avoid searching for the correct vrops, the target label is added with this pull request. This label can be used to backlink the vROps on NSX-T Alerts.


viennaa (approved):
Simple and good, thanks.

PR #197, opened 2022-02-18
target label added on NSX-T collectors

richardtief:


requirements.txt
```diff
@@ -3,8 +3,9 @@ prometheus-client
...
 PyYAML
-Flask==1.1.1
+Flask==1.1.2
 gevent
 cffi
 beautifulsoup4==4.9.3
-lxml
\ No newline at end of file
+lxml
+itsdangerous==1.1.0
```
viennaa, line 11: This is not being used, is it?

richardtief: Look at the failing tests above. This version looks like a needed dependency.

richardtief: At least for the PyTest.

viennaa: Ok, implicit dependency through Flask, I see.

viennaa (approved):
Let's try this again.

PR #205, opened 2022-03-25
move from atlas config to http service endpoint

richardtief:
The new atlas source is `http://atlas:8080/service_discovery/netbox`


viennaa:
We should actually be supporting both, file and endpoint. Merging this is blocking everyone using the exporter and not having service endpoint to query.

Issue #206, opened 2022-03-28
Possibility to configure vROps target list without webservice


viennaa:
#205 enabled the atlas service endpoint to be queried and disabled the atlas config file.

We need to have a possibility to fall back to a non-k8s-way of providing vrops targets to the exporter.

Putting a simple list into [Inventory config](https://github.com/sapcc/vrops-exporter/blob/master/tests/inventory_config.yaml) could be the solution. If given list length is > 0, we take the targets from the inventory. If it's empty or not defined at all, we try to use the native web service.

PR #207, opened 2022-03-30
inventory_config for target configuration

viennaa:
solves #206


richardtief (changes requested):
Thanks for implementing another option to provide vrops targets. I just saw a small issue.

richardtief (approved):
Thanks. Good to go.

PR #212, opened 2022-04-21
Avoidance of multiple saving of alert definitions

richardtief:
The alert definitions are a huge JSON blob that was retrieved and stored for each vrops instance and also for each alert collector. This consumed a lot of nonsensical memory.

### DIFF
--- a/collectors/AlertCollector.py
+++ b/collectors/AlertCollector.py
@@ -12,6 +12,7 @@ class AlertCollector(BaseCollector):
         super().__init__()
         self.resourcekind = list()
         self.adapterkind = list()
+        self.alert_entry_cache = dict()
 
     def get_resource_uuids(self):
         raise NotImplementedError("Please Implement this method")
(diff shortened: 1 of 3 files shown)


collectors/AlertCollector.py
```diff
@@ -73,7 +66,7 @@ def collect(self):
 
     def generate_alert_label_values(self, alert):
         alert_labels = dict()
-        alert_entry = self.alertdefinitions.get(alert.get('alertDefinitionId', {}), {})
+        alert_entry = self.get_alertdefinition(alert.get('alertDefinitionId', {}))
```
viennaa, line 69: This would add an additional call for each alert we're getting back from the `get_alerts` call.

It is a tradeoff: memory vs api load.

I'd rather suggest to implement a cache to keep the API load minimal and not have all of the definitions stored: 
1. look if there alertdefinition is locally available
2. take it, if it is not there, fetch it and take it

viennaa (commented):
This was actually pretty hidden and needed to be tackled 👍 . I'd suggest to implement a more api load friendly approach. See inline.

collectors/AlertCollector.py
```diff
@@ -72,12 +69,16 @@ def collect(self):
         yield alert_metric
 
     def generate_alert_label_values(self, alert):
+        alert_id = alert.get('alertDefinitionId', {})
         alert_labels = dict()
-        alert_entry = self.alertdefinitions.get(alert.get('alertDefinitionId', {}), {})
+        alert_entry = self.get_alertdefinition(
+            alert_id) if alert_id not in self.alert_entry_cache else self.alert_entry_cache.get(alert_id)
```
viennaa, line 75:
```suggestion
        alert_entry = self.alert_entry_cache.get(alert_id, self.get_alertdefinition(alert_id))
```

Simplifying

viennaa: `self.alert_entry_cache[alert_id]` needs to be filled here with alert_entry here to have something in there.

viennaa (approved):

viennaa:
Nice!

viennaa (approved):
Even better

PR #216, opened 2022-05-17
Exposing vrops targets from inventory

richardtief:
Exposing vrops targets from inventory helps to figure out when a vrops is not responding to requests at all.


BaseCollector.py
```diff
@@ -185,6 +185,10 @@ def get_target_tokens(self):
         self.target_tokens = self.do_request(url="http://" + os.environ['INVENTORY'] + "/target_tokens")
         return self.target_tokens
 
+    def get_vrops_list(self):
+        self.vrops_list = self.do_request(url="http://" + os.environ['INVENTORY'] + "/vrops_list")
+        return self.vrops_list
```
viennaa, line 190: This should be local only. You're always calling the function instead of the object itself.

```suggestion
        vrops_list = self.do_request(url="http://" + os.environ['INVENTORY'] + "/vrops_list")
        return vrops_list
```

viennaa (changes requested):
Good idea, using the tokens to get the vrops list in the first place was a little off anyway.

Just something small to consider.

viennaa (approved):

PR #227, opened 2022-06-20
Atlas http endpoint error handling and probe metric added

richardtief:
This PR fixes the issue where the inventory builder hangs when requesting vrops targets from atlas http sd endpoint. Error handling is now targeted at various problems that may occur and runs regardless of a successful request. Only the first run is terminated with code 0 because the `vrops_list` is not initially populated. In addition, a probe metric is supplied by the `InventoryCollector` to monitor the atlas http sd endpoint.

### DIFF
--- a/InventoryBuilder.py
+++ b/InventoryBuilder.py
@@ -179,9 +179,9 @@ class InventoryBuilder:
                     urllib3.exceptions.NewConnectionError) as err:
                 logger.error(f'Failed to establish a connection to: {self.atlas_path}, retrying in {self.sleep}s', err)
                 if not self.vrops_list:
-                    logger.critical(f'No targets found to start inventory collection, exit 0')
-                    sys.exit(0)
-                logger.info(f'continue with the old vrops_list')
+                    logger.critical(f'No targets found to start InventoryBuilder, exit 1')
+                    sys.exit(1)
+                logger.info(f'Continue with the old vrops_list')
                 return
 
             try:
@@ -189,9 +189,9 @@ class InventoryBuilder:
             except simplejson.errors.JSONDecodeError:
                 logger.error(f'Invalid json data in atlas netbox http response, Content-Type: "{response.headers.get("Content-Type")}"')
                 if not self.vrops_list:
-                    logger.critical(f'No targets found to start inventory collection, exit 0')
-                    sys.exit(0)
-                logger.info(f'continue with the old vrops_list')
+                    logger.critical(f'No targets found to start InventoryBuilder, exit 1')
+                    sys.exit(1)
+                logger.info(f'Continue with the old vrops_list')
                 return
 
             netbox_json = response.json()
--- a/InventoryBuilder.py
+++ b/InventoryBuilder.py
@@ -189,7 +189,7 @@ class InventoryBuilder:
             except simplejson.errors.JSONDecodeError:
                 logger.error(f'Invalid json data in atlas netbox http response, Content-Type: "{response.headers.get("Content-Type")}"')
                 if not self.vrops_list:
-                    logger.critical(f'No targets found to start InventoryBuilder, exit 1')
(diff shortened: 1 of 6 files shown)


InventoryBuilder.py
```diff
@@ -172,13 +174,29 @@ def get_vrops(self):
...
-                                   target['labels']['job'] == "vrops"]
+                self.response_codes['atlas'][self.atlas_path] = response.status_code
             except (urllib3.exceptions.MaxRetryError, requests.exceptions.ConnectionError,
-                    urllib3.exceptions.NewConnectionError):
-                logger.error(f'Failed to establish a connection to: {self.atlas_path}, retrying in {self.sleep}s')
+                    urllib3.exceptions.NewConnectionError) as err:
+                logger.error(f'Failed to establish a connection to: {self.atlas_path}, retrying in {self.sleep}s', err)
+                if not self.vrops_list:
+                    logger.critical(f'No targets found to start inventory collection, exit 0')
+                    sys.exit(0)
```
viennaa, line 183:
```suggestion
                    sys.exit(1)
```

Since this is not wanted behaviour I'd rather see it failing instead of gracefully quitting. This way k8s would also recognise, that something is wrong.

InventoryBuilder.py
```diff
@@ -172,13 +174,29 @@ def get_vrops(self):
...
                 logger.info(f'continue with the old vrops_list')
+                return
+
+            try:
+                response.json()
+            except simplejson.errors.JSONDecodeError:
+                logger.error(f'Invalid json data in atlas netbox http response, Content-Type: "{response.headers.get("Content-Type")}"')
+                if not self.vrops_list:
+                    logger.critical(f'No targets found to start inventory collection, exit 0')
+                    sys.exit(0)
```
viennaa, line 193: Same as above
```suggestion
                    sys.exit(1)
```

viennaa (commented):
Thanks for providing this litte helpful change. Just a minor thing with the return code to discuss.

InventoryBuilder.py
```diff
@@ -172,13 +174,29 @@ def get_vrops(self):
...
+                    sys.exit(1)
+                logger.info(f'Continue with the old vrops_list')
+                return
+
+            try:
+                response.json()
+            except simplejson.errors.JSONDecodeError:
+                logger.error(f'Invalid json data in atlas netbox http response, Content-Type: "{response.headers.get("Content-Type")}"')
+                if not self.vrops_list:
+                    logger.critical(f'No targets found to start InventoryBuilder, exit 1')
```
viennaa, line 192:
```suggestion
                    logger.critical(f'No targets found to start InventoryBuilder, bailing out')
```

InventoryBuilder.py
```diff
@@ -172,13 +174,29 @@ def get_vrops(self):
...
-                                   target['labels']['job'] == "vrops"]
+                self.response_codes['atlas'][self.atlas_path] = response.status_code
             except (urllib3.exceptions.MaxRetryError, requests.exceptions.ConnectionError,
-                    urllib3.exceptions.NewConnectionError):
-                logger.error(f'Failed to establish a connection to: {self.atlas_path}, retrying in {self.sleep}s')
-                logger.info(f'continue with the old vrops_list')
+                    urllib3.exceptions.NewConnectionError) as err:
+                logger.error(f'Failed to establish a connection to: {self.atlas_path}, retrying in {self.sleep}s', err)
+                if not self.vrops_list:
+                    logger.critical(f'No targets found to start InventoryBuilder, exit 1')
```
viennaa, line 182:
```suggestion
                    logger.critical(f'No targets found to start InventoryBuilder, bailing out')
```

viennaa (changes requested):
Thanks for providing this litte helpful change. Just a minor thing with the return code to discuss.

viennaa (approved):
Good to go!

PR #234, opened 2022-09-27
fixed amount_resources() to skip not-working targets

### LINKED ISSUE #233: Vrops-inventory-exporter dies with unavailable VRops
When the InventoryBuilder faces an unavailable VRops the exporter will not handle it properly.

`File "/vrops-exporter/collectors/InventoryCollector.py", line 31, in collect
    for gauge_metric in self.amount_inventory_resources(target):
  File "/vrops-exporter/collectors/InventoryCollector.py", line 43, in amount_inventory_resources
    for resourcekind, amount in self.get_amount_resources()[target].items():
KeyError: 'vrops-url.custom.domain.whatever'
`

Should be fixed here:
https://github.com/sapcc/vrops-exporter/blob/384e9d19e4ff216c1b8f934f3b84a4bb51961de9/collectors/InventoryCollector.py#L41-L48

`get_amount_resources()` does not have all targets from `get_vrops_list()`. Target needs to be removed from that query.

richardtief:
fixes #233


exporter.py
```diff
@@ -62,7 +63,7 @@ def parse_params(logger):
     if options.target:
         os.environ['TARGET'] = options.target
     if not options.target:
-        target = get_targets(options.inventory)[0]
+        target = random.choice(get_targets(options.inventory))
```
viennaa, line 66: What's the reason behind randomising the target query order?

viennaa (commented):
Thanks for directly taking this, just one small question.

exporter.py
richardtief: I stumbled across this by entering some targets with the unreachable vrops mgmt under `vrops_list[0]`. So the exporter was not started with a working destination. Since this option is for testing purposes only and we want to make sure it is not bound to a static target, I randomly selected it.

viennaa (approved):
Let's go

PR #236, opened 2022-09-29
fix hierarchy setting for vcops and sddc

richardtief:
- Expected storage metrics from test corrected
- Fix hierarchy setting for vcops and sddc request


viennaa (approved):
Could ship as is. One thing which would make it a little nicer though, is changing `h_dept` to `h_depth`.
At first glance, I wasn't really sure if it was a typo or if that var is defined without `h`.

PR #240, opened 2022-11-02
strike `resource_uuid` label from vm metrics

richardtief:
Reducing the complexity of virtual machine metrics as labels consume the most memory in Prometheus


viennaa (approved):
Memory--

PR #244, opened 2023-01-27
inventory per target

richardtief:


inventory.py
```diff
@@ -80,16 +74,13 @@ def parse_params(logger):
     if "INVENTORY_CONFIG" not in os.environ and not options.config:
         logger.error('Cannot start, please specify inventory config with ENV or -m')
         sys.exit(1)
-    if "ATLAS" not in os.environ and not options.atlas:
-        vrops_list = yaml_read(os.environ['INVENTORY_CONFIG']).get('vrops_targets')
-        if not vrops_list:
-            logger.error('Cannot start, please declare vrops_targets in inventory config or ATLAS path with ENV or -a')
-            sys.exit(1)
+    if "TARGET" not in os.environ and not options.target:
+        logger.error('Cannot start, please specify TARGET with ENV or -a')
```
viennaa, line 78:
```suggestion
    if "TARGET" not in os.environ and not options.target:
        logger.error('Cannot start, please specify TARGET with ENV or -t')
```

viennaa (commented):
Looks good now, thanks for fixing the test. We should consider updating the README before merging this change. All the atlas related notes need to leave. Since this is a breaking change to how it was working before we should indicate it right at the beginning of the README, in case other people are updating their code too.

viennaa (approved):
Thanks for updating the readme, looks complete now.

PR #247, opened 2023-02-09
catch sigterm in collect

### LINKED ISSUE #245: Capturing the `SIGTERM` signal in Kubernetes to finish `collect()` method.
vrops-exporters handles a lot of data in the `collect()` method that they query from the vrops API. This can take a while to complete. To not interrupt the eventual restarts during the collection cycle (maintenance controller, rolling out new versions, etc.), the termination signal should be caught and delayed to complete the collect method. This can reduce gaps in Prometheus metrics.

viennaa:
The collector should finish its collect()-run when receiving a SIGTERM. This is mainly used for a use case in kubernetes, where the given pod can be killed which would interrupt the current collect-run. Should be combined with terminationGracePeriodSeconds to avoid k8s waiting forever. A sane value would be something along 120 which is plenty of time to finish gracefully.

Fixes #245

### DIFF
--- a/collectors/PropertiesCollector.py
+++ b/collectors/PropertiesCollector.py
@@ -13,6 +13,9 @@ class PropertiesCollector(BaseCollector):
         raise NotImplementedError("Please Implement this method")
 
     def collect(self):
+        if self.collect_running and self.am_i_killed:
+            return
+        self.collect_running = True
         logger.info(f'{self.name} starts with collecting the metrics')
 
         token = self.get_target_tokens()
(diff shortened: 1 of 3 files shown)


collectors/PropertiesCollector.py
```diff
@@ -13,6 +13,9 @@ def get_labels(self, resource_id: str, project_ids: list):
         raise NotImplementedError("Please Implement this method")
 
     def collect(self):
+        if self.collect_running and self.am_i_killed:
```
richardtief, line 16: `self.collect_running` is always `false` at this point, isn't it? So `self.am_i_killed` should be enough to bail out early. 
```suggestion
        if self.am_i_killed:
```

richardtief (changes requested):
Good to see this implemented. One little remark.

collectors/PropertiesCollector.py
viennaa: That's correct, thanks.

richardtief (approved):
Thanks.

PR #248, opened 2023-02-10
sigterm for multiple collectors

viennaa:
previous integration would only work if a single collector is registered. Multiple collectors would not be aware of the others being killed. Steering moved to exporter.py, as this is the only place where we are aware of the other instantiated collectors.

### DIFF
--- a/exporter.py
+++ b/exporter.py
@@ -6,10 +6,13 @@ import importlib
 import requests
 import logging
 import random
+import signal
 from prometheus_client import start_http_server
 from prometheus_client.core import REGISTRY
 from optparse import OptionParser
 from tools.helper import yaml_read
+
+collector_globals = list()
 
 
 def default_collectors():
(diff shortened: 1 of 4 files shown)


exporter.py
```diff
@@ -109,4 +124,5 @@ def initialize_collector_by_name(class_name, logger):
     logger = logging.getLogger('vrops-exporter')
     options = parse_params(logger)
     collectors = list(map(lambda c: initialize_collector_by_name(c, logger), options.collectors))
+    collector_globals = collectors
```
richardtief, line 127: You can simply make the `collector` variable global without creating a new variable. Then it doesn't have to be passed through everywhere.
```suggestion
    global collectors
    collectors = list(map(lambda c: initialize_collector_by_name(c, logger), options.collectors))
```

richardtief (commented):
Great, works. A small note about handling of the global variable.

exporter.py
viennaa: Good catch, since it involves the change of all previous occurrences I'm going to put this in the next commit.

richardtief (approved):
Thanks. Good to go!

PR #251, opened 2023-02-22
timeout for get_resources increased

richardtief:


tools/Vrops.py
```diff
@@ -64,10 +64,10 @@ def get_adapter(self, target: str, token: str, adapterkind: str) -> (list, int):
                                     timeout=30)
         except requests.exceptions.ReadTimeout as e:
             logger.error(f'Request to {url} timed out. Error: {e}')
-            return adapter, 504
+            return adapter, 504, 0
```
viennaa, line 67: Why don't we use the real time here? Or we return the timeout value, since it should be identical.

tools/Vrops.py
```diff
@@ -178,19 +178,21 @@ def get_resources(self, target: str,
                         "name": resource["resource"]["resourceKey"]["name"],
                         "uuid": resource["resource"]["identifier"],
                         "resourcekind": resourcekind,
-                        "parent": resource.get("relatedResources", )[0],
+                        "parent": resource.get("relatedResources", [])[0],
```
viennaa, line 181: Is this the bugfix for the mentioned error in the PR?

tools/Vrops.py
```diff
@@ -178,19 +178,21 @@ def get_resources(self, target: str,
...
                         "internal_name": internal_name,
                         "instance_uuid": instance_uuid
                     })
+                    if not resource.get("relatedResources"):
+                        logger.warning(f'No parent relation returned for {resource["resource"]["resourceKey"]["name"]}; resourcekind: {resourcekind}; target: {target}.')
```
viennaa, line 186: 👍

tools/Vrops.py
```diff
@@ -242,30 +244,32 @@ def get_vms(self, target, token, parent_uuids, vcenter_uuid, query_specs):
...
+                vm_chunks, http_code, response_time = self.get_resources(target, token, adapterkind="VMWARE",
                                                                      resourcekinds=[resourcekind],
                                                                      uuids=uuid_list, query_specs=q_specs)
                 vms.extend(vm_chunks)
-                api_responding.append(api_chunk_responding)
+                http_codes.append(http_code)
+                response_times.append(response_time)
             logger.debug(f'Number of VMs collected: {len(vms)}')
-            return vms, max(api_responding)
+            return vms, max(http_codes), max(response_times)
```
viennaa, line 272: Is `max(response_times)` really a meaningful value here? Technically it will show the longest needed time to get a chunk, most likely the biggest. In the end it would be interesting to see, how long getting all VMs took. I'd rather sum it up here than taking the maximum.

viennaa (commented):
Thanks for adding the response times, looking forward to the results in prod. Two little things to talk about are inline.

tools/Vrops.py
richardtief: At this point, there is no response object. The request aborts early.

tools/Vrops.py
richardtief: That sounds like a good point. I will change that.

tools/Vrops.py
viennaa: Is it exited with the 30seconds timeout or somewhen in between?

viennaa (approved):
Good to go 👍

PR #253, opened 2023-02-27
Seperating API from builder and SIGTERM

viennaa:
it doesn't belong together and should have been separated from the beginning already
SIGTERM handling added as a prerequisite for upcoming features


InventoryBuilder.py
```diff
@@ -1,25 +1,17 @@
-import urllib3.exceptions
-from flask import Flask
-from gevent.pywsgi import WSGIServer
-from threading import Thread
 from tools.Vrops import Vrops
 from tools.helper import yaml_read
 from collections import defaultdict
+from tools.InventoryApi import InventoryApi
```
richardtief, line 4: I'm not sure if `InventoryApi` is a tool. I would prefer to group both inventory parts in one folder. e.g. `/inventory`

richardtief (commented):
Looks good. Just one little remark and one fix which I pushed directly.

InventoryBuilder.py
viennaa: Valid point, it deserves its own directory now.

richardtief (approved):
Great. Thanks!

PR #260, opened 2023-06-21
Change default value

kevin-fischer:
- Returns "_none_" for the **metric_value** if the key is not found

### DIFF
--- a/collectors/PropertiesCollector.py
+++ b/collectors/PropertiesCollector.py
@@ -58,10 +58,7 @@ class PropertiesCollector(BaseCollector):
                 values_received.add(statkey)
 
                 metric_data = value_entry.get('data', [False])[0]
-                metric_value = value_entry.get('values', [False])[0]
-
-                if metric_data == "summary|customTag:cinder_state|customTagValue" and not metric_value:
-                    metric_value = 'none'
+                metric_value = value_entry.get('values', ['none'])[0]
 
                 if statkey in metrics:
                     # enum metrics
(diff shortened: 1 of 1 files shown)


collectors/PropertiesCollector.py
```diff
@@ -60,6 +60,9 @@ def collect(self):
                 metric_data = value_entry.get('data', [False])[0]
                 metric_value = value_entry.get('values', [False])[0]
 
+                if metric_data == "summary|customTag:cinder_state|customTagValue" and not metric_value:
+                    metric_value = 'none'
+
```
richardtief, line 65: Two aspects: 

we do not build specifics into the code, it should be as generic as possible. Otherwise, in the long run, such workarounds will become a pain in the neck.

Reading this, I don't think that `metric_data` will ever fulfill the value of the statkey. Can you elaborate on that? 

In general, I would look at line 60-61, since there an empty string coming back from the API is evaluated as `False` and we want `"none"`.

richardtief (changes requested):
Thanks for providing this feature. See comments inline.

collectors/PropertiesCollector.py
kevin-fischer: Closing this pull request as I have opened a new PR to address this issue by implementing a workaround --> https://github.com/sapcc/helm-charts/pull/5117

kevin-fischer:
Closing this pull request as I have opened a new PR to address this issue by implementing a workaround --> https://github.com/sapcc/helm-charts/pull/5117

PR #266, opened 2024-02-21
feat:export complex nested values as a metric

richardtief:
There are metric keys at the vrops API that return complex data types. In most cases, they contain valuable information that only makes sense when mapped to several dedicated metrics. 

Here is an attempt to implement it as generic as possible. If there is a known key with a complex data structure for a collector, then this case must be implemented specifically. Several keys per collector are possible.

Internal vm names are mapped to the name.

Input:
```
{"rules": [{"class': 'com.integrien.alive.common.affinity.VMStorageAntiAffinityRule', 'name': 'nsxt-controller-anti-affinity', 'type': 'SDRS_VM_ANTI_AFFINITY', 'virtualMachines': ['vm-46814', 'vm-46812', 'vm-46813'], 'valid': True}]}
```

Output:
```yaml
vrops_storagepod_config_sdrsconfig_vmstorageantiaffinityrules{
      datacenter="datacenter-1",
      rule_name="nsxt-controller-anti-affinity",
      rule_type="SDRS_VM_ANTI_AFFINITY",
      rules="1/1",
      storagepod="SDRS_MGMT_DDDD",
      valid="true",
      vcenter="vcenter-dddd",
      virtualmachines="['nsx-ctl-1-dddd', 'nsx-ctl-3-dddd', 'nsx-ctl-2-dddd']"
      } 1.0

### CODE REVIEW

File: collectors/SDRSPropertiesCollector.py [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+    def unlock_nested_values(self, statkey, metric_value):
+
+        match statkey:
+            case "config|sdrsconfig|vmStorageAntiAffinityRules":
+                return self.config_sdrsconfig_vmStorageAntiAffinityRules(metric_value)
+
+    def config_sdrsconfig_vmStorageAntiAffinityRules(self, metric_value):
+
+        metric_value = json.loads(metric_value)
+        rules = metric_value.get("rules") if metric_value.get("rules") else []
```
viennaa, line 31:
```suggestion
        rules = metric_value.get("rules", [])
```

collectors/SDRSPropertiesCollector.py
```diff
@@ -0,0 +1,60 @@
...
+
+        amount_rules = len(rules) if rules else 0
```
viennaa, line 33: no need to have the if condition, len will return 0 if rules is []
```suggestion
        amount_rules = len(rules)
```

collectors/SDRSPropertiesCollector.py
```diff
@@ -0,0 +1,60 @@
...
+
+        rule_labels = ['rule', 'rule_name', 'rule_type', 'valid', 'virtualmachine']
+        rule_label_values = []
+
+        value = 0
+
+        for i, rule in enumerate(rules):
+            mapped_vms = self.vm_mapping_helper(rule.get('virtualMachines'))
+            value = len(mapped_vms)
```
viennaa, line 42: just used once, would directly return this function since it is self explanatory 
```suggestion
```

collectors/SDRSPropertiesCollector.py
```diff
@@ -0,0 +1,60 @@
...
+            for vm in mapped_vms:
+                rule_label_values.append([
+                    f'{i+1}/{amount_rules}',
+                    rule.get('name'),
+                    rule.get('type'),
+                    str(rule.get('valid')).lower(),
+                    vm
+                    ])
+        return rule_labels, rule_label_values, value
```
viennaa, line 51: see L42
```suggestion
        return rule_labels, rule_label_values, len(mapped_vms)
```

collectors/PropertiesCollector.py
```diff
@@ -60,6 +65,18 @@ def collect(self):
                 metric_data = value_entry.get('data', [False])[0]
                 metric_value = value_entry.get('values', [False])[0]
 
+                if statkey in self.nested_value_metric_keys:
+                    add_labels, add_label_value_list, add_value = self.unlock_nested_values(statkey, metric_value)
+                    if add_labels[0] not in metrics[statkey]['gauge']._labelnames:
```
viennaa, line 70: This is... hacky :>.

Add the `label_names` before L35, when `self.generate_metrics(label_names=self.label_names)` is called. It should not rely on the internal data structure. You can check on `self.nested_value_metric_keys` and extend the list. You've got to sort out, which metric is affected. Maybe it is even needed to do it in the `BaseCollector` and add a flag to pass through.

collectors/SDRSPropertiesCollector.py
```diff
@@ -0,0 +1,60 @@
...
+
+        amount_rules = len(rules) if rules else 0
+
+        rule_labels = ['rule', 'rule_name', 'rule_type', 'valid', 'virtualmachine']
+        rule_label_values = []
+
+        value = 0
+
+        for i, rule in enumerate(rules):
+            mapped_vms = self.vm_mapping_helper(rule.get('virtualMachines'))
```
viennaa, line 41: Can this be empty? 
```suggestion
            mapped_vms = self.vm_mapping_helper(rule.get('virtualMachines'), [])
```

viennaa (changes requested):
Finally some real code review again, I like it!

Added some refinements and would like to avoid the use of the internal data structure of GaugeMetricFamily.

collectors/SDRSPropertiesCollector.py
richardtief: That wouldn't work.  The `rules` key is always present and filled with `None`, which cannot be iterated. The following would be useful:
```suggestion
        rules = metric_value.get("rules", []) if metric_value.get("rules") else []
```

richardtief: This will work? It's both `false`, no?

collectors/PropertiesCollector.py
richardtief: True, but we also use the internal variable further down in the collectors. We can wrap this in a function so that it doesn't appear so prominently in the code.

richardtief: Latest commit reflects this idea.

collectors/SDRSPropertiesCollector.py
viennaa: Yes this works. If the left hand side is `None`, the right hand side is returned.

```
>>> metric_value
{'rules': None}
>>> foobar = metric_value.get("rules") or []
>>> foobar
[]
>>> metric_value = {"rules": ["a","b","c"]}
>>> foobar = metric_value.get("rules") or []
>>> foobar
['a', 'b', 'c']
```

collectors/SDRSPropertiesCollector.py
```diff
@@ -47,7 +45,7 @@ def config_sdrsconfig_vmStorageAntiAffinityRules(self, metric_value):
                     str(rule.get('valid')).lower(),
                     vm
                     ])
-        return rule_labels, rule_label_values, len(mapped_vms)
+        return rule_labels, rule_label_values, 1
```
viennaa, line 48: `1` is static, because this is no _real_ value here?

collectors/PropertiesCollector.py
viennaa: Yeah I like this more. The use of `._labelnames` happened during the rewrite a very while back. Wasn't aware of this anymore.

collectors/SDRSPropertiesCollector.py
richardtief: At the end it should be `sum` -able by promql. `sum by (rule_name)` should lead to 3.

viennaa: ok

viennaa (approved):
Cool, let's go!

Testing for this is still on the old branch, that's why it is failing.

kevin-fischer:
Great job on this!
TY for your work, @richardtief and @viennaa.

